PR #93, opened 2019-06-17
Added REST API documentation for Java Authentication

mrm9084:


docs/REST/authentication.md
```diff
@@ -289,3 +289,74 @@ static class HttpContentExtensions
     }
 }
 ```
+### Java
+```java
+public class ConfigHttpClient {
+
+    public CloseableHttpResponse execute(HttpUriRequest request, Date date, String credential, String secret) {
```
zhenlan: Can we obtain the `date` in the `sign` function? A user doesn't have to care about it.

docs/REST/authentication.md
```diff
@@ -289,3 +289,74 @@ static class HttpContentExtensions
...
+
+        // Signature
+        String methodName = request.getRequestLine().getMethod().toUpperCase();
+        URIBuilder uri = new URIBuilder(request.getRequestLine().getUri());
+        String scheme = uri.getScheme() + "://";
+        String requestPath = uri.toString().substring(scheme.length()).substring(uri.getHost().length());
+        String host = new URIBuilder(request.getRequestLine().getUri()).getHost();
+        String toSign = String.format("%s\n%s\n%s;%s;%s", methodName, requestPath, requestTime, host, contentHash);
+
+        // Signature
```
zhenlan: The segment above is also commented with `// Signature`. Can we merge them?

docs/REST/authentication.md
```diff
@@ -289,3 +289,74 @@ static class HttpContentExtensions
...
+
+    private static String buildContentHash(HttpUriRequest request) {
+        String content = "";
+        if (request instanceof HttpEntityEnclosingRequest) {
+            try {
+                StringWriter writer = new StringWriter();
+                IOUtils.copy(((HttpEntityEnclosingRequest) request).getEntity().getContent(), writer,
+                        StandardCharsets.UTF_8);
+
+                return writer.toString();
```
zhenlan: I may miss something, but I don't see the hash is computed for the content here.

docs/REST/authentication.md
```diff
@@ -289,3 +289,74 @@ static class HttpContentExtensions
...
+            }
+            finally {
+                try {
+                    ((HttpEntityEnclosingRequest) request).getEntity().getContent().close();
+                }
+                catch (IOException e) {
+                    LOGGER.trace("Failed to close the input stream.", e);
```
zhenlan: Is the `LOGGER` a standard type in Java? I am wondering if we should swallow the exception here or just let it bubble up.

docs/REST/authentication.md
```diff
@@ -289,3 +289,74 @@ static class HttpContentExtensions
...
+                }
+            }
+        }
+
+        byte[] digest = new DigestUtils(SHA_256).digest(content);
```
zhenlan: `content` is only initialized as `""` at the beginning. Will this always produces the same result?

zhenlan:
Thanks, @mrm9084. What are the environments this code will work? Does it need Spring Boot, Spring Framework or just any Java environment? What are the environments you tried?

docs/REST/authentication.md
mrm9084: Yeah, that return should be a content = writer.toString().

docs/REST/authentication.md
mrm9084: I wasn't sure how to handle it, it is not a standard output. I could just let it be thrown, that way they have to deal with it on there end, or they can make a change here.

docs/REST/authentication.md
mrm9084: See your previous comment. line 346 needs to be updated.

mrm9084:
@zhenlan This code snippet just uses apache commons 2.5 and apache http components, so as long as you include those two libraries you are good.

docs/REST/authentication.md
```diff
@@ -289,3 +289,68 @@ static class HttpContentExtensions
...
+    public CloseableHttpResponse execute(HttpUriRequest request, String credential, String secret)
+            throws ClientProtocolException, IOException, UnsupportedOperationException, URISyntaxException {
+        Map<String, String> authHeaders = sign(request, credential, secret);
+        authHeaders.forEach(request::setHeader);
+
+        return httpClient.execute(request);
+    }
+
+    private static Map<String, String> sign(HttpUriRequest request, String credential,
+            String secret) throws URISyntaxException, UnsupportedOperationException, IOException {
```
zhenlan: NIT: Just for easier to read, can we have `String secret)` on the previous line?

docs/REST/authentication.md
```diff
@@ -289,3 +289,68 @@ static class HttpContentExtensions
...
+        String authorization = String.format("HMAC-SHA256 Credential=%s, SignedHeaders=%s, Signature=%s",
+                credential, signedHeaders, signature);
+        headers.put("Authorization", authorization);
+
+        headers.put(HttpHeaders.USER_AGENT, USER_AGENT);
+
+        return headers;
+    }
+
+    private static String buildContentHash(HttpUriRequest request) throws UnsupportedOperationException, IOException {
```
zhenlan: Why `UnsupportedOperationException`?

docs/REST/authentication.md
```diff
@@ -289,3 +289,68 @@ static class HttpContentExtensions
...
+        // Compose headers
+        Map<String, String> headers = new HashMap<>();
+        headers.put("x-ms-date", requestTime);
+        headers.put("x-ms-content-sha256", contentHash);
+
+        String authorization = String.format("HMAC-SHA256 Credential=%s, SignedHeaders=%s, Signature=%s",
+                credential, signedHeaders, signature);
+        headers.put("Authorization", authorization);
+
+        headers.put(HttpHeaders.USER_AGENT, USER_AGENT);
```
zhenlan: Why do we need this line for signing? A user can always add a user-agent header if they want.

zhenlan:
> @zhenlan Zhenlan Wang FTE This code snippet just uses apache commons 2.5 and apache http components, so as long as you include those two libraries you are good.

Great. Can we include this be part of your code snippet or in a comment at the beginning?

docs/REST/authentication.md
```diff
@@ -289,3 +289,68 @@ static class HttpContentExtensions
     }
 }
 ```
+### Java
+```java
+public class ConfigHttpClient {
```
zhenlan: NIT: No need the class here.

docs/REST/authentication.md
```diff
@@ -289,3 +289,62 @@ static class HttpContentExtensions
...
+public CloseableHttpResponse execute(HttpUriRequest request, String credential, String secret)
+        throws IOException, URISyntaxException {
+    Map<String, String> authHeaders = sign(request, credential, secret);
+    authHeaders.forEach(request::setHeader);
+
+    return httpClient.execute(request);
+}
+
+private static Map<String, String> sign(HttpUriRequest request, String credential, String secret) 
+        throws URISyntaxException, IOException {
```
zhenlan: I am wondering if the input is an object of `HttpUriRequest` (instead of a uri string), is the `URISyntaxException` possible?

mrm9084: The `URISyntaxException` comes from lines 312 and 315 with the use of `URIBuilder`, where it thinks it is possible to throw an exception as it takes in the URI as a String and not as a URI object. As for `HttpUriRequest`, which in the case of the provider is actually a `HttpGet`, it does not validate that the Uri is valid when creating the `HttpGet`.

docs/REST/authentication.md
```diff
@@ -289,3 +289,62 @@ static class HttpContentExtensions
     }
 }
 ```
+### Java
+```java
+public CloseableHttpResponse execute(HttpUriRequest request, String credential, String secret)
+        throws IOException, URISyntaxException {
+    Map<String, String> authHeaders = sign(request, credential, secret);
+    authHeaders.forEach(request::setHeader);
```
MSGaryWang: >authHeaders.forEach(request::setHeader); [](start = 4, length = 40)

We had passed request into sign(), shouldn't we just set header inside it? Or call the method getAuthHeader()?

mrm9084: This code is based off how the provider does it. It's mainly for readability where the `authHeaders` are generated, then attached to the request.

MSGaryWang: sign() function name doesn't clearly describe the purpose because it didn't sign it yet but generate the header for you. People need to read the code to know this.

---
In reply to: [297318652](https://github.com/Azure/AppConfiguration/pull/93#discussion_r297318652) [](ancestors = 297318652)

mrm9084: I updated it from sign() to generateHeader()

Issue #81, opened 2019-06-05
Any example about how to create credential and secret?


newsunwu:
I am investigating how to use AppConfiguration in my prototype service.
The REST API has great example about using AppConfiguration API, link is below. But it does not 
mention how to get the credential and secret. Are the credential and secret retrieved from 
AppConfiguration service or it is generated on the fly. 

Could document include an example how to get credential and secret?

Thank you very much for your great help

Best Regards!
Wu Sun

jimmyca15:
@newsunwu the credential and secret are obtained from the Azure App Configuration access keys. If using the portal to view the access keys then only the connection string is shown, but the connection string contains the secret and credential.

Endpoint={endpoint};Id={credential};Secret={secret}

I see that the authentication doc is a bit out of sync with the fact that credential is also referred to as the secret 'Id'. Will need to update it.

zhenlan:
@mlacouture let's also update our portal UI to list the `id` and `secret` separately in addition to the connection string.

mlacouture:
the portal part is done and will be available in the next prod deployment!

mlacouture:
this is now fixed in prod.

zhenlan:
@jimmyca15 do you think anything needs to be changed in the doc?

jimmyca15:
I think it could be a little unclear so I updated it with pr #94.

jimmyca15:
This was resolved in portal and docs.

PR #94, opened 2019-06-18
Updated rest auth docs to detail how to obtain 'Credential' and 'Secret'.

jimmyca15:
This is in response to #81.

### DIFF
--- a/docs/REST/authentication.md
+++ b/docs/REST/authentication.md
@@ -6,6 +6,7 @@ All HTTP requests must be authenticated using the **HMAC-SHA256** authentication
 - **Credential** - \<Access Key ID\>
 - **Secret** - base64 decoded Access Key Value. ``base64_decode(<Access Key Value>)``
 
+These values must be obtainted from the Azure App Configuration instance. This can be done using the [Azure Portal](https://portal.azure.com) or the [Azure CLI](https://docs.microsoft.com/en-us/cli/azure/?view=azure-cli-latest). Given a connection string obtained from either one of these sources, one can extract the credential and secret by parsing their respective segments: Endpoint={endpoint};Id={**Credential**};Secret={**Secret**}
 
 Provide each request with all HTTP headers required for Authentication. The minimum required are:
 


docs/REST/authentication.md
```diff
@@ -6,6 +6,7 @@ All HTTP requests must be authenticated using the **HMAC-SHA256** authentication
 - **Credential** - \<Access Key ID\>
 - **Secret** - base64 decoded Access Key Value. ``base64_decode(<Access Key Value>)``
 
+These values must be obtainted from the Azure App Configuration instance. This can be done using the [Azure Portal](https://portal.azure.com) or the [Azure CLI](https://docs.microsoft.com/en-us/cli/azure/?view=azure-cli-latest). Given a connection string obtained from either one of these sources, one can extract the credential and secret by parsing their respective segments: Endpoint={endpoint};Id={**Credential**};Secret={**Secret**}
```
zhenlan: The connection string should be opaque to users. We shouldn't ask users to parse it. This is why both of our portal and CLI list `id` and `secret` separately in addition to the `connection string`. So I will remove the second sentence in the above paragraph. There could be confusion about `credential` and `id`. That's why our portal showed it as `Id (credential)`. We probably can clarify here they are the same thing.

zhenlan: Just realized our CLI calls the `secret` as `value`. Is this naming from RP? If so, we can clarify it here too.
```
C:\Windows\System32>az appconfig credential list -n ConfigStoreDemo
[
  {
    "connectionString": "xxxxx",
    "id": "xxxxx",
    "lastModified": "2019-02-25T18:52:32+00:00",
    "name": "Primary",
    "readOnly": false,
    "value": "xxxxx"
  },
 ...
]
```

jimmyca15: I'll remove the connection string sentence. Although it is an interesting that if we don't document it then people who want to develop their own client can't really leverage it. In other words develop their own AppConfigClient that takes connection string. 

Yes, value comes from the RP. The CLI shows the raw response. I'll clarify.

PR #103, opened 2019-07-03
Add golang sample code

MSGaryWang:
For authentication sample code piece

### DIFF
--- a/docs/REST/authentication.md
+++ b/docs/REST/authentication.md
@@ -292,7 +292,7 @@ static class HttpContentExtensions
 ```
 ### Java
 ```java
-public CloseableHttpResponse execute(HttpUriRequest request, String credential, String secret)
+public CloseableHttpResponse signRequest(HttpUriRequest request, String credential, String secret)
         throws IOException, URISyntaxException {
     Map<String, String> authHeaders = generateHeader(request, credential, secret);
     authHeaders.forEach(request::setHeader);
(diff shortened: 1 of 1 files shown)


docs/REST/authentication.md
```diff
@@ -348,4 +348,47 @@ private static String buildContentHash(HttpUriRequest request) throws IOExceptio
...
+}
+
+func signRequest(content string, key []byte) string {
+    hmac := hmac.New(sha256.New, key)
+    hmac.Write([]byte(content))
+    return base64.StdEncoding.EncodeToString(hmac.Sum(nil))
+}
+
+// Setup the auth header for accessing Azure AppConfiguration service
+func PrepareAuthHeader(verb string, host string, pathAndQuery string, body string, id string, secret string, req *http.Request) {
```
jimmyca15: >verb string, host string, pathAndQuery string, body string [](start = 23, length = 58)

Is it common to pass around everything when working with HTTP request in go?

I would expect just passing in (req, id, secret) like in java or C#. Then getting req.Method, req.Host, req.URL, and req.GetBody() to prepare the authentication. With all the current parameters it kind of seems like the httprequest should be generated in the method rather than passed in.

docs/REST/authentication.md
```diff
@@ -348,4 +348,47 @@ private static String buildContentHash(HttpUriRequest request) throws IOExceptio
...
+
+    timestamp := time.Now().UTC().Format(http.TimeFormat)
+    contentHash := getContentHashBase64(body)
+    signature := signRequest(fmt.Sprintf("%s\n%s\n%s;%s;%s", strings.ToUpper(verb), pathAndQuery, timestamp, host, contentHash), key)
+
+    req.Header.Set("x-ms-content-sha256", contentHash)
+    req.Header.Set("x-ms-date", timestamp)
+    req.Header.Set("Authorization", "HMAC-SHA256 Credential="+id+", SignedHeaders=x-ms-date;host;x-ms-content-sha256, Signature="+signature)
+    req.Header.Set("Content-Type", "application/vnd.microsoft.appconfig.kv+json")
+}
```
jimmyca15: This is not related to authentication.

docs/REST/authentication.md
```diff
@@ -348,4 +348,47 @@ private static String buildContentHash(HttpUriRequest request) throws IOExceptio
...
+// Setup the auth header for accessing Azure AppConfiguration service
+func PrepareAuthHeader(verb string, host string, pathAndQuery string, body string, id string, secret string, req *http.Request) {
+    key, err := base64.StdEncoding.DecodeString(secret)
+    if err != nil {
+        fmt.Printf("err = %s \n", err)
+    }
+
+    timestamp := time.Now().UTC().Format(http.TimeFormat)
+    contentHash := getContentHashBase64(body)
+    signature := signRequest(fmt.Sprintf("%s\n%s\n%s;%s;%s", strings.ToUpper(verb), pathAndQuery, timestamp, host, contentHash), key)
```
zhenlan: Just like other examples, it's better to split this to separate lines to have a `stringToSign`, which will be easier to read.

docs/REST/authentication.md
```diff
@@ -348,4 +348,47 @@ private static String buildContentHash(HttpUriRequest request) throws IOExceptio
...
+    "time"
+)
+
+func getContentHashBase64(content string) string {
+    hasher := sha256.New()
+    hasher.Write([]byte(content))
+    return base64.StdEncoding.EncodeToString(hasher.Sum(nil))
+}
+
+func signRequest(content string, key []byte) string {
```
zhenlan: Technically, this is not signing a request. It's just calculating hmac of a given string. Maybe `getHmac` is better?

Can these helper functions be put to the end of the code or do they have to be declared before called in go?

docs/REST/authentication.md
zhenlan: BTW, it will be easier to read to just call this function `Sign` or `SignRequest`. It will be consistent with other examples.

zhenlan: Do you mind to change the `execute` in Java example to `SignRequest` too?

PR #108, opened 2019-07-11
Update MicrosoftExtensionsConfigurationAzureAppConfiguration.md

abhilasharora:
Add release notes for Microsoft.Extensions.Configuration.AzureAppConfiguration v2.0.0-preview-009200001-1437

### DIFF
--- a/releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
+++ b/releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
@@ -1,13 +1,12 @@
 ## Microsoft Extensions Azure App Configuration
 
 ### v2.0.0-preview-009200001-1437 - July 10, 2019
-* Added `ConfigureRefresh` method to configure dynamic refresh for configuration settings
+* Replaced watch feature with on-demand refresh mechanism
+    * Watch feature had issues that resulted in too many server requests to keep the configuration updated
+    * Refresh feature with an internal cache resolves the issues with watch feature and allows on-demand refresh
+    * Removed `Watch` and `WatchAndReloadAll` methods
+    * Added `ConfigureRefresh` method to configure dynamic refresh for configuration settings
     * Added `GetRefresher` method to retrieve an `IConfigurationRefresher` to trigger manual refresh in code
-    * Added an internal cache to limit calls to the configuration store when refresh is triggered too often
-* Removed the Watch feature to refresh configuration at regular polling intervals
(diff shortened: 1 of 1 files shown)


releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
```diff
@@ -1,5 +1,14 @@
 ## Microsoft Extensions Azure App Configuration
 
+### v2.0.0-preview-009200001-1437 - July 10, 2019
+* Added `ConfigureRefresh` method to configure dynamic refresh for configuration settings
```
jimmyca15: >Added `ConfigureRefresh` method to configure dynamic refresh for configuration settings [](start = 2, length = 87)

I'm trying to put my self in the shoes of a developer reading these release notes, and I'm thinking that it would be more helpful if rather than treating these like disjointed addition and removal of APIs we wrote it something like

* Replaced Watch capability with on-demand refresh mechanism.
  * Watch has some bad things
  * Refresh solves these bad things, and allows you to refresh on demand
* `Watch` no longer available
* Added `ConfigureRefresh`
* Added `GetRefresher`

That's an abbreviated version, but I think the structure is more helpful because the release was more of a redesign/replacement rather than adding/removing things. What do you think?

zhenlan: Agree with @jimmyca15. I was thinking about the same.
(I assume "bad things" are just placeholder :))

releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
```diff
@@ -1,5 +1,13 @@
 ## Microsoft Extensions Azure App Configuration
 
+### v2.0.0-preview-009200001-1437 - July 10, 2019
+* Replaced watch feature with on-demand refresh mechanism
+    * Watch feature had issues that resulted in too many server requests to keep the configuration updated
```
jimmyca15: I think this might be a little vague for those looking back and wondering why watch was abandoned. Also the wording makes it sound like there was an issue in `Watch` that could have just been fixed rather than the fact that it was fundamentally an undesirable design that we moved away from. I would like to be super informative for this one. How about 

"The watch mechanism had some fundamental design flaws. It was not able to be invoked on-demand. It was always running even in applications that could be considered inactive. It promoted constant polling of configuration rather than a more intelligent approach of refreshing configuration when applications are active or need to ensure freshness"

PR #151, opened 2019-09-26
Add authentication code snippet for Python

avanigupta:


docs/REST/authentication.md
```diff
@@ -407,3 +407,82 @@ func getHmac(content string, key []byte) string {
...
+### Python
+```python
+
+import base64
+import hashlib
+import hmac
+from datetime import datetime
+import six
+
+def sign_request(method, url, body, connection_string):
```
shenmuxiaosen: >connection_string [](start = 36, length = 17)

I think we need to do a little modification here. For other languages, the input is credential and secret instead of connection string. So in our case, we don't need the parse connection string helper method.

docs/REST/authentication.md
```diff
@@ -407,3 +407,58 @@ func getHmac(content string, key []byte) string {
...
+
+    if six.PY2:
+        content_digest = hashlib.sha256(bytes(body)).digest()
+    else:
+        content_digest = hashlib.sha256(bytes(body, 'utf-8')).digest()
+
+    content_hash = base64.b64encode(content_digest).decode('utf-8')
+
+    # Signed Headers
+    signed_headers = "x-ms-date;host;x-ms-content-sha256"  # Semicolon separated header names
```
shenmuxiaosen: >; [](start = 36, length = 1)

replace ";" with "&" and fix the comment

shenmuxiaosen:
LGTM. Thanks Avani!

PR #155, opened 2019-09-30
Add PowerShell authentication sample

et13:
Add PowerShell authentication sample

### DIFF
--- a/docs/REST/authentication.md
+++ b/docs/REST/authentication.md
@@ -464,8 +464,9 @@ def sign_request(host,
 ### PowerShell
 ```PowerShell
 function Sign-Request(
-    [System.Uri] $uri,
+    [string] $serverHost,
     [string] $method,      # GET, PUT, POST, DELETE
+    [string] $url,         # path+query
     [string] $body,        # request body
     [string] $credential,  # access key id
     [string] $secret       # access key value (base64 encoded)
@@ -478,8 +479,8 @@ function Sign-Request(
     $signedHeaders = "x-ms-date;host;x-ms-content-sha256";  # Semicolon separated header names
 
     $stringToSign = $verb + "`n" +
-                    $uri.PathAndQuery + "`n" +
-                    $utcNow + ";" + $uri.Authority + ";" + $contentHash  # Semicolon separated signedHeaders values
+                    $url + "`n" +
+                    $utcNow + ";" + $serverHost + ";" + $contentHash  # Semicolon separated signedHeaders values
 
     $signature = Compute-HMACSHA256Hash $secret $stringToSign
  
@@ -527,6 +528,6 @@ $body = $null
 $credential = "<Credential>"
 $secret = "<Secret>"
 
-$headers = Sign-Request $uri $method $body $credential $secret
+$headers = Sign-Request $uri.Authority $method $uri.PathAndQuery $body $credential $secret
 Invoke-RestMethod -Uri $uri -Method $method -Headers $headers -Body $body
 ```
--- a/docs/REST/authentication.md
+++ b/docs/REST/authentication.md
@@ -223,7 +223,7 @@ using (var client = new HttpClient())
 {
     var request = new HttpRequestMessage()
     {
-        RequestUri = new Uri("http://example.azconfig.io/kv"),
+        RequestUri = new Uri("https://{config store name}.azconfig.io/kv"),
         Method = HttpMethod.Get
     };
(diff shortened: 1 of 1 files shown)


docs/REST/authentication.md
```diff
@@ -461,3 +461,72 @@ def sign_request(host,
...
+        "x-ms-date" = $utcNow;
+        "x-ms-content-sha256" = $contentHash;
+        "Authorization" = "HMAC-SHA256 Credential=" + $credential + "&SignedHeaders=" + $signedHeaders + "&Signature=" + $signature
+    }
+}
+
+# Stop if any error occurs
+$ErrorActionPreference = "Stop"
+
+$uri = [System.Uri]::new("http://example.azconfig.io/kv")
```
jimmyca15: http**s**.

Also, it may be better to use a placeholder store name. someone can create a store named example.

`"https://{config store name}.azconfig.io/kv"`

docs/REST/authentication.md
```diff
@@ -461,3 +461,72 @@ def sign_request(host,
...
+$verb = "GET"
```
jimmyca15: Can we use `Method` which corresponds to the `-Method` parameter of `Invoke-RestMethod`.

jimmyca15:
Minor suggestions. Looks great otherwise.

docs/REST/authentication.md
```diff
@@ -461,3 +461,72 @@ def sign_request(host,
...
+    $hmac = [System.Security.Cryptography.HMACSHA256]::new([Convert]::FromBase64String($secret))
+    try {
+        return [Convert]::ToBase64String($hmac.ComputeHash([Text.Encoding]::ASCII.GetBytes($content)))
+    }
+    finally {
+        $hmac.Dispose()
+    }
+}
+
+function Sign-Request(
```
drago-draganov: I think we should make this look more explicit? 
Also, if possible, place it on the top of the snippet. It's about signing request to begin with.

```sh
Sign-Request(host, 
                      method,      // GET, PUT, POST, DELETE
                     url,         // path+query
                     body,        // request body (undefined of none)
                     credential,  // access key id
                     secret)      // access key value (base64 encoded)
```sh

File: docs/REST/authentication.md (hunk 2 above)
Comment (drago-draganov) [reply docs/REST/authentication.md]: Can we also include the ```-Method``` in Invoke-RestMethod?

File: docs/REST/authentication.md (hunk 1 above)
Comment (et13) [reply docs/REST/authentication.md]: Sure, I made the change here and in the C# sample (which had the same example url)

File: docs/REST/authentication.md (hunk 2 above)
Comment (et13) [reply docs/REST/authentication.md]: Yes, I renamed this to method and passed it into Invoke-RestMethod (-Method)

File: docs/REST/authentication.md (hunk 3 above)
Comment (et13) [reply docs/REST/authentication.md]: Sure, I moved Sign-Request to the top.  Could you clarify what you mean by making this look more explicit?  Is it about having separate paramters for host and url (path+query)?  I had seen that the javascript/python samples had separate parameters while the C#/Java/Golang didn't (host/path/query pulled out of the uri/url in the Request object).  I had gone with the Uri parameter because it is also the input into Invoke-RestMethod.

Comment (drago-draganov) [reply docs/REST/authentication.md]: This is a document that indeed focus on HMAC authentication.

The PS code snippet is to demonstrate how the HMAC is computed. It's not SDK or complete solution for PS scenarios. Therefore I think we should focus on the details and primitives outlined by the document. 

C#/Java/Golang snippets all work based on their respective HttpRequest objects, not on raw Uri.

Comment (et13) [reply docs/REST/authentication.md]: Ok, I updated Sign-Request to accept the host and url (path/query) like the javascript and python samples.

Note:  I added a qualifier to the parameter name `host` due to a name conflict with the PowerShell automatic variable also named `host`.

File: docs/REST/authentication.md [hunk 4]
```diff
@@ -461,3 +461,73 @@ def sign_request(host,
         "Authorization": "HMAC-SHA256 Credential=" + credential + "&SignedHeaders=" + signed_headers + "&Signature=" + signature
     }
 ```
+### PowerShell
+```PowerShell
+function Sign-Request(
+    [string] $serverHost,
```
jimmyca15: how about `$hostname`. serverHost host sounds a bit strange.

et13: Thanks, I renamed the param

PR #166, opened 2019-10-25
Fix the link to the authentication doc

zhenlan:


jimmyca15:
@zhenlan I need this change for a doc sync I'm about to do between our other doc set in vsts repo. I'm going to go ahead and merge it.

PR #167, opened 2019-10-25
Added version doc.

jimmyca15:
Added version doc.


jimmyca15:
There are changes missed in this. I'm going to abandon this and resubmit.

PR #168, opened 2019-10-25
Added versioning doc.

jimmyca15:
Added versioning doc.

### DIFF
--- a/docs/REST/authentication.md
+++ b/docs/REST/authentication.md
@@ -92,7 +92,7 @@ _String-To-Sign=_
 string-To-Sign=
             "GET" + '\n' +                                                                                      // VERB
             "/kv?fields=*" + '\n' +                                                                             // path_and_query
-            "Fri, 11 May 2018 18:48:36 GMT;example.azconfig.io;47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU="    // signed_headers_values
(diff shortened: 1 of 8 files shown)


jimmyca15:
@zhenlan this PR has all the missing pieces for versioning and latest changes in the internal doc set.

docs/REST/authentication.md
```diff
@@ -21,7 +21,7 @@ Provide each request with all HTTP headers required for Authentication. The mini
 
 **Example:**
 ```
-Host: example.azconfig.io
+Host: {myconfig}.azconfig.io
```
zhenlan: Since this is an example, maybe we should use contoso.azconfig.io? In line 95, there is another *example.azconfig.io* that we probably should keep consistent with.

zhenlan:
Can we add the link of the versioning doc to the front page as well?
https://github.com/Azure/AppConfiguration#rest-api-reference

docs/REST/authentication.md
jimmyca15: We did this in the docs because those names aren't reserved. Looks like line 95 wasn't missed.

jimmyca15:
@zhenlan Yes it certainly should be there. Thanks, updated.

zhenlan:
Is the update to the readme.md missed from the PR?

jimmyca15:
@zhenlan Looks like it didn't get included in my last commit. Added it.

PR #182, opened 2019-11-05
Update MicrosoftAzureAppConfigurationAspNetCore.md

abhilasharora:


jimmyca15:
Shouldn't both release notes be in the same PR?

abhilasharora:
> Shouldn't both release notes be in the same PR?

I created the pull requests using GitHub web interface without enlisting the branch, so the only option was to create separate PRs. I will merge them in appropriate order.

jimmyca15:
@abhilasharora oh, I see. That makes sense.

PR #187, opened 2019-11-11
Add an example of Azure Function

zhenlan:


examples/DotNetCore/AzureFunction/FunctionApp/Function1.cs
```diff
@@ -0,0 +1,58 @@
...
+using Microsoft.Azure.WebJobs.Extensions.Http;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Configuration.AzureAppConfiguration;
+using Microsoft.Extensions.Logging;
+
+namespace FunctionApp
+{
+    public static class Function1
+    {
+        static readonly MyConfig _myConfig = new MyConfig();
```
jimmyca15: >new MyConfig(); [](start = 45, length = 15)

If we do this can we just add 

```
        public IConfiguration Configuration { set; get; }
        public IConfigurationRefresher ConfigurationRefresher { set; get; }
```

as static members and then use a static constructor to initialize them.

jimmyca15: Then we can avoid the extra type.

---
In reply to: [346051128](https://github.com/Azure/AppConfiguration/pull/187#discussion_r346051128) [](ancestors = 346051128)

jimmyca15: >static readonly [](start = 8, length = 15)

Can you add a comment to explain why static.

examples/DotNetCore/AzureFunction/FunctionApp/Function1.cs
```diff
@@ -0,0 +1,58 @@
...
+            // Configuration will be refreshed if the cache is expired and the 'Sentinel' key is modified.
+            // Remove the 'await' operator if the configuration is preferred to be refreshed without blocking.
+            await _myConfig.ConfigurationRefresher.Refresh();
+
+            string keyName = "TestApp:Settings:Message";
+            string message = _myConfig.Configuration[keyName];
+            
+            return message != null
+                ? (ActionResult)new OkObjectResult(message)
+                : new BadRequestObjectResult($"Please pass a message from App Configuration with key name '{keyName}'");
```
jimmyca15: >pass a message from App Configuration with key name [](start = 54, length = 51)

The code is already set up to pass the message. IF they are seeing this then I would expect they just forgot to create the key-value in the app configuration. We could be more explicit and say "Please create a key-value with the key '{keyName}' in App Configuration."

examples/DotNetCore/AzureFunction/FunctionApp/Function1.cs
zhenlan: Good idea.

zhenlan: Yes, absolutely.

examples/DotNetCore/AzureFunction/FunctionApp/Function1.cs
zhenlan: Sounds good.

examples/DotNetCore/AzureFunction/FunctionApp/Function1.cs
```diff
@@ -0,0 +1,58 @@
...
+using Microsoft.Azure.WebJobs.Extensions.Http;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Configuration.AzureAppConfiguration;
+using Microsoft.Extensions.Logging;
+
+namespace FunctionApp
+{
+    public static class Function1
+    {
+        // Use the static modifier to create a singleton instance of Configuraion. This avoids
```
jimmyca15: Configura**t**ion

zhenlan: Thanks!

PR #200, opened 2019-11-28
Adjust table size and fix some links

zhenlan:
- Break module name to multiple lines
- Fix the sample links of .net feature management


README.md
```diff
@@ -36,25 +36,25 @@ The following reference pages are available to describe the Azure App Configurat
...
+Microsoft.Extensions.Configuration. AzureAppConfiguration | .Net Standard | [![NuGet](https://img.shields.io/nuget/v/Microsoft.Extensions.Configuration.AzureAppConfiguration.svg?color=blue)](https://www.nuget.org/packages/Microsoft.Azure.AppConfiguration.AspNetCore/) | [Sample](https://github.com/Azure/AppConfiguration/tree/master/examples/DotNetCore) | [Release Notes](https://github.com/Azure/AppConfiguration/blob/master/releaseNotes/MicrosoftAzureAppConfigurationAspNetCore.md)
+Microsoft.Azure.AppConfiguration.<br/>AspNetCore | ASP&#46;NET Core | [![NuGet](https://img.shields.io/nuget/v/Microsoft.Azure.AppConfiguration.AspNetCore.svg?color=blue)](https://www.nuget.org/packages/Microsoft.Azure.AppConfiguration.AspNetCore/) | [Sample](https://github.com/Azure/AppConfiguration/tree/master/examples/DotNetCore) | [Release Notes](https://github.com/Azure/AppConfiguration/blob/master/releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md)
```
jimmyca15: I'm not sure we should be adding custom line breaks for styling. The styling outcome is relative to screen size, trying to work around it may not work as expected.

zhenlan: True. I hope there is a better way. But the current one looks very bad - the version is hardly readable. The fix didn't put any hard coded size. It only split the name into multiple lines. They will scale as normal.

zhenlan: Alright. I removed the version column and put it to the same column of the module name. This avoided splitting the module name to multiple lines but still gave enough space for the version to be readable. Does this look better?

BTW, I also added a one-sentence description of what each section is about.

README.md
```diff
@@ -36,30 +36,36 @@ The following reference pages are available to describe the Azure App Configurat
...
+[Microsoft.Configuration.ConfigurationBuilders.AzureAppConfiguration](https://github.com/aspnet/MicrosoftConfigurationBuilders/tree/master/src/AzureAppConfig)<br/>[![NuGet](https://img.shields.io/nuget/v/Microsoft.Configuration.ConfigurationBuilders.AzureAppConfiguration.svg?color=blue)](https://www.nuget.org/packages/Microsoft.Configuration.ConfigurationBuilders.AzureAppConfiguration/) | .NET Framework |  |
+[spring-cloud-azure-appconfiguration-config](https://github.com/microsoft/spring-cloud-azure/tree/master/spring-cloud-azure-appconfiguration-confige)<br/>[![Maven Central](https://img.shields.io/maven-central/v/com.microsoft.azure/spring-cloud-azure-appconfiguration-config.svg?color=blue)](https://search.maven.org/artifact/com.microsoft.azure/spring-cloud-azure-appconfiguration-config) | Spring Boot | [Sample](https://github.com/microsoft/spring-cloud-azure/tree/master/spring-cloud-azure-samples/azure-appconfiguration-sample) | [Release Notes](https://github.com/Azure/AppConfiguration/blob/master/releaseNotes/SpringCloudAzureAppConfigurationConfig.md)
```
mrm9084: The GitHub url has an e at end. Should be https://github.com/microsoft/spring-cloud-azure/tree/master/spring-cloud-azure-appconfiguration-config

zhenlan: Good catch. Fixed.

README.md
jimmyca15: This new way looks very good. It also is good usage of <br/>.

PR #219, opened 2019-12-10
Spring Release Notes M6

mrm9084:


releaseNotes/SpringCloudAzureFeatureManagement.md
```diff
@@ -25,4 +32,4 @@
 
 [package_web]: https://mvnrepository.com/artifact/com.microsoft.azure/spring-cloud-azure-feature-management-web
 [samples_web]: https://github.com/microsoft/spring-cloud-azure/tree/master/spring-cloud-azure-samples/feature-management-web-sample
-[source_code_web]: https://github.com/microsoft/spring-cloud-azure/tree/master/spring-cloud-azure-feature-management-web
\ No newline at end of file
+[source_code_web]: https://github.com/microsoft/spring-cloud-azure/tree/master/spring-cloud-azure-feature-management-web
```
zhenlan: Please add an empty line to the end of the file.

releaseNotes/SpringCloudAzureFeatureManagement.md
```diff
@@ -1,9 +1,16 @@
...
+
 [Source code][source_code] | [Package (Maven)][package] | [Samples][samples] | [Product documentation][docs]
 
 # spring-cloud-azure-feature-management-web
+
 [Source code web ][source_code_web] | [Package (Maven) web][package_web] | [Samples web][samples_web] | [Product documentation][docs]
 
+## 1.1.0.M6 - December 09, 2019
+
+* Added support for '.' in key values from files.
```
zhenlan: Not sure what this means... do we support file name to have a dot or do we support the key name to have a dot?

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,16 @@
 
 [Source code][source_code] | [Package (Maven)][package] | [Product documentation][docs] | [Samples][samples]
 
+## 1.1.0.M6 - December 09, 2019
+
+* Added support for AAD, users can now [Azure TokenCredentials][token_credentials]. Token Credential support allows setting Azure Client ID, Azure Tenant ID, and Azure Client Secret through Environment Variables.
```
lisaguthrie: "Users can now..."? Is there a word missing here? Maybe "Users can now use..." or "Users can now access..."?

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,16 @@
...
+* Adds support for providing Token Credentials to the provider using the TokenCredentialProvider.
+* Fix Bug where off feature filters would be processed incorrectly resulting in the feature filters still running.
```
lisaguthrie: "off feature filters" is awkward -- maybe "disabled feature filters" instead?

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,16 @@
...
+
+### Breaking Change
+
+* The Role Contributor is no longer used. Two new roles have been added App Configuration Data Owner and App Configuration Data Reader. The member added to the role only needs App Configuration Data Reader for the provider.
```
lisaguthrie: Suggest adding a colon to the second sentence - "Two new roles have been added: App Configuration Data Owner..."

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,50 @@
...
 
+## 1.1.0.M6 - December 09, 2019
+
+* Updated support for AAD. System Assigned identity previously required; config store name, and the Contributor role. Now it requires; config store config store name, [client id][client_id], and the App Configuration Data Reader role.
+
+```properties
+spring.cloud.azure.appconfiguration.managed-identity.client-id=[client id]
+spring.cloud.azure.appconfiguration.stores[0].name=[config store name]
+```
+
```
zhenlan: How about this?

* Updated managed identity support for both system-assigned managed identity and user-assigned managed identity.

  **Before**
The managed identity is assigned *Contributor* role to the App Configuration instance.

  **After**
  - The system-assigned managed identity also requires the client-id to be set in addition to the config store name, for example,
    ```properties
    spring.cloud.azure.appconfiguration.managed-identity.client-id=[client id]
    spring.cloud.azure.appconfiguration.stores[0].name=[config store name]
    ```
  - The managed identity should be assigned *App Configuration Data Reader* role to the App Configuration instance.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,50 @@
...
+* Updated support for AAD. User Assigned identity previously required; config store name, client id, object id, and the Contributor role. Now it required; config store name, client id, and the App Configuration Data Reader role.
```
zhenlan: This can be removed as it's already covered in the first bullet point already.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,50 @@
...
+    TokenCredential buildCredential() {
+            return new UsernamePasswordCredentialBuilder()
+                    .clientId("<YOUR_CLIENT_ID>")
+                    .username("<YOUR_USERNAME>")
+                    .password("<YOUR_PASSWORD>")
+                    .build();
+    }
+
+}
+```
```
zhenlan: How about this?

* Added support of generic Azure Active Directory (AAD) authentication. Users can use any of the Azure TokenCredentials from the [Azure Identity client library][token_credentials] to access the App Configuration and Key Vault. 
  1. Set `spring.cloud.azure.appconfiguration.stores[0].name` to your App Configuration store name.
  2. Add a `@Bean` of `TokenCredentialProvider` to your application. For example, the following code snippet used the `UsernamePasswordCredential` for the authorization with both the App Configuration and Key Vault.

        ```java
        public class MyCredentials implements TokenCredentialProvider {
            @Override
            public TokenCredential credentialForAppConfig() {
                    return buildCredential();
            }
            @Override
            public TokenCredential credentialForKeyVault() {
                    return buildCredential();
            }
            TokenCredential buildCredential() {
                    return new UsernamePasswordCredentialBuilder()
                            .clientId("<YOUR_CLIENT_ID>")
                            .username("<YOUR_USERNAME>")
                            .password("<YOUR_PASSWORD>")
                            .build();
            }
        }
        ```

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,50 @@
...
+
+```properties
+spring.cloud.azure.appconfiguration.managed-identity.client-id=[client id]
+spring.cloud.azure.appconfiguration.stores[0].name=[config store name]
+```
+
+* Updated support for AAD. User Assigned identity previously required; config store name, client id, object id, and the Contributor role. Now it required; config store name, client id, and the App Configuration Data Reader role.
+* Added the following Provider to allow users to provide any [Azure TokenCredentials][token_credentials] to the provider for authenticating with App Configuration & Key Vault. When a `TokenCredentialProvider` `@Bean` is made it will be used for authentication.
+
+**Note:** Null is valid value to be returned if only one of the authentication methods needs to be used.
```
zhenlan: I think we shouldn't fall back to the managed identity if the user code returns null. This is too magic.  If a user is already implementing a `TokenCredentialProvider`, it's simple enough for them to return a `ManagedIdentityCredential`.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,50 @@
...
+                    .clientId("<YOUR_CLIENT_ID>")
+                    .username("<YOUR_USERNAME>")
+                    .password("<YOUR_PASSWORD>")
+                    .build();
+    }
+
+}
+```
+
+* An error is now thrown when more than 1 authentication method is provided.
```
zhenlan: What exactly is the error that will be thrown?

I think we should be more specific here. How about this?

* To prevent unintentional results, a {specific error} exception will be thrown if both `spring.cloud.azure.appconfiguration.stores[0].name` and `spring.cloud.azure.appconfiguration.stores[0].connection-string` are provided.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,50 @@
...
+* Adds support for providing Token Credentials to the provider using the TokenCredentialProvider.
```
zhenlan: Remove as this is already mentioned.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,50 @@
...
+* Fix Bug where disabled feature filters would be processed incorrectly resulting in the feature filters still running.
```
zhenlan: This is feature management. This shouldn't belong here.

For the text, how about this?

* Fixed a bug where feature filters are still evaluated even if the feature flag is disabled.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -14,7 +58,9 @@
 * Added Feature Management Support.
 
 <!-- LINKS -->
-[docs]: https://docs.microsoft.com/en-us/azure/azure-app-configuration/quickstart-java-spring-app
+[client_id]: https://github.com/microsoft/spring-cloud-azure
```
zhenlan: I don't think we need this.

releaseNotes/SpringCloudAzureFeatureManagement.md
```diff
@@ -1,9 +1,15 @@
...
+
 [Source code][source_code] | [Package (Maven)][package] | [Samples][samples] | [Product documentation][docs]
 
 # spring-cloud-azure-feature-management-web
+
 [Source code web ][source_code_web] | [Package (Maven) web][package_web] | [Samples web][samples_web] | [Product documentation][docs]
 
+## 1.1.0.M6 - December 09, 2019
+
+* Fix bug where feature flag names were unable to have '.' when read from a configuration file, but valid when read from Azure.
```
zhenlan: How about this?

* Fix a bug where feature flag names were unable to contain dots (`.`) when they are read from a configuration file.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
mrm9084: @Zhenlan The idea is that they could use the Provider for Key Vault, but a connection string for App Configuration

PR #240, opened 2020-01-09
Release Notes for JavaProvider 1.2.x

mrm9084:


releaseNotes/SpringCloudAzureFeatureManagement.md
```diff
@@ -6,6 +6,33 @@
 
 [Source code web ][source_code_web] | [Package (Maven) web][package_web] | [Samples web][samples_web] | [Product documentation][docs]
 
+## 1.2.0 - January 13, 2020
+
+* Feature Management config files now match the DotNet format.
```
zhenlan: Please use Java developer-friendly schema here and not mention "match the dotnet".

releaseNotes/SpringCloudAzureFeatureManagement.md
```diff
@@ -6,6 +6,33 @@
...
+        Name: Random
+        Parameters:
+          chance: "50"
+      -
+        Name: ClientFilter
+        Parameters:
+          clientIp: 10.0.0.1
+```
+
+* The isEnabled method is now non-blocking Async and has been renamed to isEnabledAsync.
```
zhenlan: The reason for this change is to allow async feature filter. Anything the user needs to know to write an async feature filter?

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,25 @@
 
 [Source code][source_code] | [Package (Maven)][package] | [Product documentation][docs] | [Samples][samples]
 
+## 1.2.0 - January 13, 2020
+
+* A number of configuration parameters have renamed/replaced/removed. See [starter](https://github.com/microsoft/spring-cloud-azure/blob/master/spring-cloud-azure-starters/spring-cloud-starter-azure-appconfiguration-config/README.md) for more info.
```
zhenlan: Let's be specific of all the breaking changes. For a while, this could be the only document users have to use to move to this release.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,25 @@
...
+
+```properties
+spring.cloud.azure.appconfiguration.watch.delay -> spring.cloud.azure.appconfiguration.cache.expiration
+
+# Endpoint Requires URI value
+spring.cloud.azure.appconfiguration.stores[0].name -> spring.cloud.azure.appconfiguration.stores[0].endpoint
+```
+
+* AzureCloudConfigWatch is now named AzureCloudConfigRefresh. Also AzureCloudConfigRefresh is always created and can be access via dependency injection to allow control over refreshes outside of ServletRequestHandledEvents.
```
zhenlan: Is `AzureCloudConfigWatch` something a user knows about in the past? If not, we don't have to mention it.

zhenlan: BTW, is `AzureCloudConfigRefresh` a name from us? If so, shouldn't AppConfig be part of the name?

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,25 @@
...
+* System Assigned Credential no longer needs client id set.
+* TokenCredentialProvider has been split into two file AppConfigCredentialProvider and KeyVaultCredentialProvider.
+* AzureCloudConfigRefresh is now non-blocking Async using Reactor.
```
zhenlan: Can we group similar things together?

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,25 @@
...
+* Bug fix, multiple refreshes no longer occur when a system that is watching multiple stores where two or more stores are updated before the local values are updated.
+* Bug fix, Sentinels are update before refresh completes, resulting in a failed refresh not being reattempted.
```
zhenlan: "Sentinels" may not be a concept everyone knows. How about

* Fixed the bug a failed configuration refresh may not be reattempted.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,25 @@
...
+
+# Endpoint Requires URI value
+spring.cloud.azure.appconfiguration.stores[0].name -> spring.cloud.azure.appconfiguration.stores[0].endpoint
+```
+
+* AzureCloudConfigWatch is now named AzureCloudConfigRefresh. Also AzureCloudConfigRefresh is always created and can be access via dependency injection to allow control over refreshes outside of ServletRequestHandledEvents.
+* System Assigned Credential no longer needs client id set.
+* TokenCredentialProvider has been split into two file AppConfigCredentialProvider and KeyVaultCredentialProvider.
+* AzureCloudConfigRefresh is now non-blocking Async using Reactor.
+* Bug fix, multiple refreshes no longer occur when a system that is watching multiple stores where two or more stores are updated before the local values are updated.
```
zhenlan: I may not understand what you mean correctly. Is this what you are saying?

* Fixed the bug that configuration refresh may occur multiple times unnecessarily when an application loads configuration from more than one App Configuration store.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,25 @@
...
+
+```properties
+spring.cloud.azure.appconfiguration.watch.delay -> spring.cloud.azure.appconfiguration.cache.expiration
+
+# Endpoint Requires URI value
+spring.cloud.azure.appconfiguration.stores[0].name -> spring.cloud.azure.appconfiguration.stores[0].endpoint
+```
+
+* AzureCloudConfigWatch is now named AzureCloudConfigRefresh. Also AzureCloudConfigRefresh is always created and can be access via dependency injection to allow control over refreshes outside of ServletRequestHandledEvents.
+* System Assigned Credential no longer needs client id set.
```
zhenlan: * Fixed the bug that system-assigned managed identity no longer needs client id to be set.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,47 @@
 
 [Source code][source_code] | [Package (Maven)][package] | [Product documentation][docs] | [Samples][samples]
 
+## 1.1.1/1.2.1 - January 13, 2020
+
```
zhenlan: We should use full sentences and publicly known wording (instead of acronyms) for the release notes wherever we can. Whenever possible, we should also explain why we are making each change. For example, instead of this
- Watch enabled has been removed. To control the refresh rate the cache expiration configuration can be used. By default 30s.

We can have something like
- The settings below are removed as the timer-based configuration watch has been deprecated.
  ```properties
  spring.cloud.azure.appconfiguration.watch.enabled
  spring.cloud.azure.appconfiguration.watch.delay
  ```
- The configuration is refreshed based on application activities.
    - The configuration is cached and the default cache expiration time is 30 seconds. The configuration won't be refreshed until the cache expiration time window is reached. This value can be modified, for example,
  ```properties
  spring.cloud.azure.appconfiguration.cache-expiration = 60
  ```
    - In a web application, this library will signal configuration refresh automatically as long as there are incoming requests to the application (quick question, will this cause our library to depend on some web library?).
    - In a non-web application, the user needs to xxx (call `AzureCloudConfigRefresh`? is there a code snippet?) to signal configuration refresh at places where application activities occur.

There are also duplicated information. Please clean them up.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,47 @@
...
+* 4 configuration parameters have renamed/replaced/removed. See [starter](https://github.com/microsoft/spring-cloud-azure/blob/master/spring-cloud-azure-starters/spring-cloud-starter-azure-appconfiguration-config/README.md) for updated descriptions.
+
+```properties
+# The following configurations have been removed
+spring.cloud.azure.appconfiguration.watch.enabled
+spring.cloud.azure.appconfiguration.managed-identity.object-id
+
+# The following property has been renamed
+spring.cloud.azure.appconfiguration.watch.delay -> spring.cloud.azure.appconfiguration.cache.expiration
```
zhenlan: Should it be spring.cloud.azure.appconfiguration.**cache-expiration**?

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,47 @@
...
+
+# Endpoint Requires URI value
+spring.cloud.azure.appconfiguration.stores[0].name -> spring.cloud.azure.appconfiguration.stores[0].endpoint
+```
+
+* Watch enabled has been removed. To control the refresh rate the cache expiration configuration can be used. By default 30s.
+* Managed Identity doesn't need an object id any more.
```
zhenlan: Duplicated with what you had under authentication.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,47 @@
...
+* Watch delay was renamed to cache expiration to reflect new use.
+* For clarity the name configuration has been changed to endpoint. Instead of my-configstore-name use `https://my-configstore-name.azconfig.io`
+
+### Authentication
+
+* TokenCredentialProvider has been split into two file AppConfigCredentialProvider and KeyVaultCredentialProvider.
```
zhenlan: I don't think we talked about `TokenCredentialProvider` previously, so it's really the new introduction of `AppConfigCredentialProvider` and `KeyVaultCredentialProvider`.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,47 @@
...
+# The following property has been renamed
+spring.cloud.azure.appconfiguration.watch.delay -> spring.cloud.azure.appconfiguration.cache.expiration
+
+# Endpoint Requires URI value
+spring.cloud.azure.appconfiguration.stores[0].name -> spring.cloud.azure.appconfiguration.stores[0].endpoint
+```
+
+* Watch enabled has been removed. To control the refresh rate the cache expiration configuration can be used. By default 30s.
+* Managed Identity doesn't need an object id any more.
+* Watch delay was renamed to cache expiration to reflect new use.
```
zhenlan: Duplicated

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,47 @@
...
+* For clarity the name configuration has been changed to endpoint. Instead of my-configstore-name use `https://my-configstore-name.azconfig.io`
+
+### Authentication
+
+* TokenCredentialProvider has been split into two file AppConfigCredentialProvider and KeyVaultCredentialProvider.
+* Bug fix, system-assigned managed identity no longer needs client id to be set.
+
+### Refresh
+
+* AzureCloudConfigRefresh can be access via dependency injection to allow control over refreshes outside of ServletRequestHandledEvents. The refresh will only update after the cache expires based on the configured value.
```
zhenlan: accessed

releaseNotes/SpringCloudAzureFeatureManagement.md
```diff
@@ -6,6 +6,33 @@
...
 
+## 1.2.0 - January 13, 2020
+
+* Simplified Feature Management config file format.
+
+```yaml
+feature-management:
+  Beta: true
+  DarkTheme:
+    EnabledFor:
```
zhenlan: Don't we have the Java-style for this, for example, `enabled-for`?

releaseNotes/SpringCloudAzureFeatureManagement.md
```diff
@@ -6,6 +6,33 @@
...
 [Source code web ][source_code_web] | [Package (Maven) web][package_web] | [Samples web][samples_web] | [Product documentation][docs]
 
+## 1.2.0 - January 13, 2020
+
+* Simplified Feature Management config file format.
+
+```yaml
+feature-management:
+  Beta: true
+  DarkTheme:
```
zhenlan: This is for Java developers. If it makes more sense for a Java developer to use `data-theme` instead of `DarkTheme`, we should do so in our examples.

releaseNotes/SpringCloudAzureFeatureManagement.md
```diff
@@ -6,6 +6,33 @@
 
 [Source code web ][source_code_web] | [Package (Maven) web][package_web] | [Samples web][samples_web] | [Product documentation][docs]
 
+## 1.2.0 - January 13, 2020
+
+* Simplified Feature Management config file format.
```
zhenlan: What we give here is just an example. Is the formatting schema published somewhere?

releaseNotes/SpringCloudAzureFeatureManagement.md
```diff
@@ -6,6 +6,33 @@
...
+        Name: Random
+        Parameters:
+          chance: "50"
+      -
+        Name: ClientFilter
+        Parameters:
+          clientIp: 10.0.0.1
+```
+
+* The isEnabled method is now non-blocking Async and has been renamed to isEnabledAsync using Reactor.
```
zhenlan: Please quote APIs, for example, `isEnabled`.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
mrm9084: @zhenlan When this was initially changed we already had a dependency on org.springframework.web, from the shared dependency of the spring-cloud-azure-autoconfigure library. But now with use of the Azure sdk, that dependency was removed, so now this is the only thing causing us to have a web dependency. We can rethink it, we could either remove it completely, or drop it from listening to `ServletRequestHandledEvent` to `ApplicationEvent`, but that may be to often. The only other option would be to have the listener be in the sample and have the user implement it.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
mrm9084: No, before it was an object called watch with a delay Duration being set, now it is just the renamed object to cache with exiration Duration being set.

releaseNotes/SpringCloudAzureFeatureManagement.md
mrm9084: This part here is the name of the feature, so I imange it would be like this inside of App Configuration, so if they exported to a file this is what they would get. Both would work here, I just imange this being more likely.

releaseNotes/SpringCloudAzureFeatureManagement.md
mrm9084: This is running into issues as it doesn't work well supporting both data coming from both App Config and a configuration file as we have to process them differently. When coming from a file "." is not a valid character to have inside of a feature name. i.e.

```yaml
feature-management:
 feature.a: true
```

This value coming from the file is read as:

```yaml
feature-management:
  feature:
    a: true
```

But this just works coming from App Config. The bigger issue where `enabled-for` comes in is that coming from one it is read as a list of items i.e: [{Name:"Random", Parameters:[...]}, {Name:"ClientFilter", Parameters:[...]}], but the other reads them as a HashMap to fake being a list. {0:{Name:"Random", Parameters:[...]}, 1:{Name:"ClientFilter", Parameters:[...]}}.

I am looking into beter ways to solve this list hashmap difference. Which results in us using enabledFor and EnabledFor to seperate the two and allow support for enabled-for as right now it doesn't seem to allow for it to be set.

releaseNotes/SpringCloudAzureFeatureManagement.md
mrm9084: Do you mean the API for the value we return which is Mono? i.e. https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Mono.html?

releaseNotes/SpringCloudAzureFeatureManagement.md
mrm9084: We don't have a published schema for the Feature Management configuration file. I will start working on one to add on with the Azure App Configuration one.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
mrm9084: Yes

releaseNotes/SpringCloudAzureFeatureManagement.md
zhenlan: I mean use single backticks so \`isEnabled\` will look like `isEnabled`.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
zhenlan: In this case, I think we should create a separate library for the web.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,72 @@
...
+```properties
+spring.cloud.azure.appconfiguration.cache-expiration = 60
+```
+
+* Bug fix, configuration refresh occurred multiple times unnecessarily when an application loads configuration from more than one App Configuration store.
+* Bug fix, failed configuration refresh may not be reattempted.
+
+### Authentication
+
+* The property below has been removed as the previous AAD authentication has been replaced with the Azure Java SDK.
```
zhenlan: * The property below has been removed as the object ID is not needed for managed identity authentication.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,72 @@
...
+
+```properties
+spring.cloud.azure.appconfiguration.managed-identity.object-id
+```
+
+* For clarity the name configuration has been changed to endpoint. Instead of my-configstore-name use `https://my-configstore-name.azconfig.io`
```
zhenlan: * With the generic AAD support by the App Configuration, the property `name` has been renamed to `endpoint`
  **Before**
  ```properties
  spring.cloud.azure.appconfiguration.stores[0].name={my-configstore-name}
  ```
  **After**
  ```properties
  spring.cloud.azure.appconfiguration.stores[0].endpoint= https://{my-configstore-name}.azconfig.io
  ```

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,72 @@
...
+
+```properties
+spring.cloud.azure.appconfiguration.stores[0].endpoint= https://my-configstore-name.azconfig.io
+```
+
+* A new authentication method has been added allowing users to provide there own credentials. Users can create a AppConfigCredentialProvider and KeyVaultCredentialProvider @Bean to provide credentials to App Configuration for connection to Azure App Configuration and Azure Key Vault respectively. Both objects define a method that is given a uri value and returns a [TokenCredential][token_credentials].
```
zhenlan: * Users can now implement `AppConfigCredentialProvider` or `KeyVaultCredentialProvider` to use any of the `TokenCredential` types that is supported by the [Azure Identity](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/identity/azure-identity/README.md) to authenticate with App Configuration or Key Vault respectively. Please see xxx document for details.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,72 @@
...
+  * The configuration is cached and the default cache expiration time is 30 seconds. The configuration won't be refreshed until the cache expiration time window is reached. This value can be modified, see example below.
+  * In a web application, this library will signal configuration refresh automatically as long as there are incoming requests to the application. This is done by listening for ServletRequestHandledEvents.
+  * In a non-web application, the user needs to call AzureCloudConfigRefresh's `refreshConfigurations` method to signal configuration refresh at places where application activities occur. AzureCloudConfigRefresh can be accessed via dependency injection. Web based applications can also use this method, though both methods will still only update when the cache has expired. `refreshConfigurations` will return a `Future<Boolean>`, the resulting Boolean can be checked to see if a refresh event was triggered. **Note:** This does not mean the updated has been completed, only that a request was made to have the configurations updated.
+
+```properties
+spring.cloud.azure.appconfiguration.cache-expiration = 60
+```
```
zhenlan: Why don't we put this just next to the bullet point that is talking about his property?

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,72 @@
...
+
+* Added Support for Spring Boot 2.2.x Hoxton with version 1.2.1. Spring Boot 2.1.x Greenwich is still supported with 1.1.1.
+
+### Configuration
+
+* The property below has been removed as the timer-based configuration watch has been deprecated.
+
+```properties
+spring.cloud.azure.appconfiguration.watch.enabled
+```
```
zhenlan: you can add two spaces in front of this for better alignment of the indentation.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,72 @@
...
+
+* The configuration is refreshed based on application activities.
+  * The configuration is cached and the default cache expiration time is 30 seconds. The configuration won't be refreshed until the cache expiration time window is reached. This value can be modified, see example below.
+  * In a web application, this library will signal configuration refresh automatically as long as there are incoming requests to the application. This is done by listening for ServletRequestHandledEvents.
+  * In a non-web application, the user needs to call AzureCloudConfigRefresh's `refreshConfigurations` method to signal configuration refresh at places where application activities occur. AzureCloudConfigRefresh can be accessed via dependency injection. Web based applications can also use this method, though both methods will still only update when the cache has expired. `refreshConfigurations` will return a `Future<Boolean>`, the resulting Boolean can be checked to see if a refresh event was triggered. **Note:** This does not mean the updated has been completed, only that a request was made to have the configurations updated.
```
zhenlan: This does not mean the update~~d~~ has been completed...

If we are going to change the design of `AzureCloudConfigRefresh` soon, I am thinking we probably don't bother to advertise it now. Thoughts?

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,72 @@
...
+
+```properties
+spring.cloud.azure.appconfiguration.cache-expiration = 60
+```
+
+* Bug fix, configuration refresh occurred multiple times unnecessarily when an application loads configuration from more than one App Configuration store.
```
zhenlan: Fixed the bug that configuration refresh...

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,72 @@
...
+* Bug fix, failed configuration refresh may not be reattempted.
```
zhenlan: Fixed the bug that...

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,72 @@
...
+    }
+
+    TokenCredential buildCredential() {
+            return new DefaultAzureCredentialBuilder().build();
+    }
+
+}
+```
+
+* Bug fix, system-assigned managed identity no longer needs client id to be set.
```
zhenlan: Fixed the bug that...

releaseNotes/SpringCloudAzureFeatureManagement.md
zhenlan: If we are not ready to announce this, let's don't mention it this time.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
mrm9084: The sub intents don't work well with the code section between them, so I just put it afterwards.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
mrm9084: I will do this for all of them.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
mrm9084: I don't think any anything about this will change. The fail fast change don't happen until after `RefreshEvent` takes place. So if the failure happens when checking if a `RefreshEvent` needs to be called, then the user will be able to catch the error.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
zhenlan: You will need spaces in front of \`\`\`properties and \`\`\` too.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
zhenlan: I am talking about the name change of `AzureCloudConfigRefresh`.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
zhenlan: It worked fine for me. See https://github.com/Azure/AppConfiguration/pull/240#discussion_r366731435.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,79 @@
...
+  spring.cloud.azure.appconfiguration.stores[0].name={my-configstore-name}
+  ```
+
+#### After
+
+  ```properties
+  spring.cloud.azure.appconfiguration.stores[0].endpoint= https://{my-configstore-name}.azconfig.io
+  ```
+
+* Users can now implement `AppConfigCredentialProvider` and/or `KeyBaultCredentialProvider` to use any of the `TokenCredential` authentication types that are supported by [Azure Identity][token_credentials] to authenticate with App Configuration or Key Vault respectively. Please see [Starter](https://github.com/mrm9084/spring-cloud-azure/tree/master/spring-cloud-azure-starters/spring-cloud-starter-azure-appconfiguration-config) for more details.
```
zhenlan: > ... to authenticate with App Configuration or Key Vault...

Technically, this is authorization instead of authentication (the user is already authenticated with AAD when they obtain the token). That's why I used the term "access App Configuration or Key Vault...".

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -2,6 +2,79 @@
...
+      @Override
+      public TokenCredential credentialForKeyVault(String uri) {
+              return buildCredential();
+      }
+
+      TokenCredential buildCredential() {
+              return new DefaultAzureCredentialBuilder().build();
+      }
+  }
+  ```
```
zhenlan: Since you already point to your full document for more details, I don't think you need to duplicate this code snippet here in the release notes.

PR #266, opened 2020-02-19
Update Azure Functions example to use dependency injection.

zhenlan:
This way the single instance of IConfiguration from the App Configuration will be shared among multiple Azure Functions within the same Function App.

Feature management is also added as part of the example.


examples/DotNetCore/AzureFunction/FunctionApp/Startup.cs
```diff
@@ -0,0 +1,41 @@
...
+            // Load configuration from Azure App Configuration
+            ConfigurationBuilder configurationBuilder = new ConfigurationBuilder();
+            configurationBuilder.AddAzureAppConfiguration(options =>
+            {
+                options.Connect(Environment.GetEnvironmentVariable("ConnectionString"))
+                       // Load all keys that start with `TestApp:`
+                       .Select("TestApp:*")
+                       // Configure to reload configuration if the registered key is modified
+                       .ConfigureRefresh(refreshOptions =>
+                            refreshOptions.Register("TestApp:Settings:Message", refreshAll: true)
```
abhilasharora: Since we know the `refreshAll` parameter is most useful when it is tracking a sentinel key, we can consider changing the key name registered with refreshAll: true from `TestApp:Settings:Message` to something like `TestApp:Settings:ConfigVersion` or `TestApp:Settings:Sentinel`.

zhenlan: I was debating myself between simplicity and best practice. You have a good point about `refreshAll`. Sure, I will change back to use the sentinel key.

examples/DotNetCore/AzureFunction/FunctionApp/ShowMessage.cs
```diff
@@ -29,7 +29,7 @@ public ShowMessage(IOptionsSnapshot<Settings> settings, IConfigurationRefresher
 
             // Signal to refresh the configuration if the registered key(s) is modified.
             // This will be a no-op if the cache expiration time window is not reached.
-            // The configuration is refreshed asynchronously.
+            // The configuration is refreshed asynchronously without blocking the execustion of the current function.
```
abhilasharora: Typo : `execustion` -> `execution`

zhenlan: Thanks!

Issue #255, opened 2020-01-27
ConfigurationBuilder Hangs With Bad AAC Connection String


jdroman2302:
Hi team,
  
  My apologies if this is a known issue, but I did not see anything related in the open issues list.  
It seems that after setting up the ConfigurationBuilder with AddAzureAppConfiguration **and** specifying a bad connection string, when you call ConfigurationBuilder.Build(), it actually hangs.  I would have expected some kind of http exception after a reasonable timeout, but it just seems to hang indefinitely.
  I have found this behavior in the preview 1152 build, using both NetFx 4.72 and NetCore 3.1 console clients.  Also passing the "optional" argument as both true and false, neither seems to have an impact on this issue.
  
  I would really appreciate if someone could look at this.  We are hoping to build our multitenant architecture around AAC, but if our apps cannot start and then do not get meaningful exceptions, that presents a real problem.

  Thanks!

abhilasharora:
@jdroman2302 Thanks for reporting the issue. In case the connection string is invalid due to incorrect format, or invalid values of `Id` or `Secret`, an exception is thrown promptly with the appropriate error message. However, in case the name of the configuration store in the connection string is incorrect, then the provider has an internal retry mechanism, where it continues to connect to the configuration store until all retry attempts have been exhausted.

Did the connection string used in your case had a non-existent configuration store name? I tried that scenario and I do see an exception after waiting for around 10 minutes. Also, do you see the application to be stuck for a longer period?

jdroman2302:
@abhilasharora Yes in my case I was testing with a connection string that was valid except for the first segment (the store name).  I only waited about 3-5 minutes and assumed it was permanently stuck.  That's good to know about the 10 minute retry/limit, but obviously not good to experience that behavior on startup.  Is there any way to control the behavior in 1152?  If not, can this be addressed in an upcoming release?

jdroman2302:
@abhilasharora @zhenlan any update on this?  I'm concerned about moving forward with the provider if it waits 10 minutes before reporting the store is down.  If it will at least be fixed in the GA version 1, that will make me feel better.

abhilasharora:
@jdroman2302 This issue will be fixed in the next release of the package. We plan to reduce the time it takes to fail in such cases to a reasonable value, with an additional option for the user to control the number of retry attempts for such failures.

jdroman2302:
Thanks guys, will look forward to that.

zhenlan:
[...]

PR #270, opened 2020-02-20
Update MicrosoftExtensionsConfigurationAzureAppConfiguration.md

abhilasharora:


releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
```diff
@@ -1,6 +1,15 @@
 ## Microsoft.Extensions.Configuration.AzureAppConfiguration
 ### [Package (NuGet)](https://www.nuget.org/packages/Microsoft.Extensions.Configuration.AzureAppConfiguration)
 
+### 3.0.0 - February 19, 2020
+* Added the following method to allow users to override `ConfigurationClientOptions`. These can be used to update the default retry options or configure a proxy for connecting to Azure App Configuration.
+
+   ````csharp
+   public AzureAppConfigurationOptions ConfigureClientOptions(Action<ConfigurationClientOptions> options)
+   ````
+* Added `IConfigurationRefresher.TryRefreshAsync` method to make it easier to trigger configuration refresh without crashing the application due to transient errors. Also, renamed the `Refresh` method to `RefreshAsync`.
```
jimmyca15: Can we put the rename into a separate bullet.

releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
```diff
@@ -1,6 +1,15 @@
...
+* Reduced the number of retry attempts made for failed requests on initial configuration load from 12 to 2. This prevents manual refresh from blocking the application for long periods of time.
```
jimmyca15: Can we shorten it a bit? Something like "Reduced maximum number of retries when query App Configuration to prevent blocking the application for long periods of time during refresh."

releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
```diff
@@ -1,6 +1,15 @@
 ## Microsoft.Extensions.Configuration.AzureAppConfiguration
 ### [Package (NuGet)](https://www.nuget.org/packages/Microsoft.Extensions.Configuration.AzureAppConfiguration)
 
+### 3.0.0 - February 19, 2020
+* Added the following method to allow users to override `ConfigurationClientOptions`. These can be used to update the default retry options or configure a proxy for connecting to Azure App Configuration.
```
jimmyca15: The usage of the word 'or' makes it seem like those are the only two possible use cases. We could change to something like "This enables customization on the underlying App Configuration client including modifying retry options and configuring proxy settings."

zhenlan: In addition to what Jimmy said, you can add a link to the [issue](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/106) now.

releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
zhenlan: Add a link to the issue too.
https://github.com/Azure/AppConfiguration/issues/255

releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
zhenlan: The exception will not necessarily crash the app. It depends on how users call it. How about this?

"Added `IConfigurationRefresher.TryRefreshAsync` method, which will not throw exceptions on transient errors during configuration refresh. [Azure/AppConfiguration-DotnetProvider/issues/113]"

I agree to separate the rename one to make it more obvious.

PR #294, opened 2020-03-11
Update Restful doc throttling

Yiming-Jia:


docs/REST/throttling.md
```diff
@@ -5,9 +5,13 @@ Configuration stores have limits on the requests that they may serve. Any reques
 Throttling is divided into different quota policies:
 
 **TotalRequests** - total amount of requests/sec
-**TotalWrites** - total amount of write requests/second
+		
+**WriteRequests** - total amount of write and delete requests/second
```
abhilasharora: We can replace `write` by `update` to avoid confusion since a delete request is also a form of write request.

**WriteRequests** - total amount of update and delete requests/second

docs/REST/throttling.md
```diff
@@ -5,9 +5,13 @@ Configuration stores have limits on the requests that they may serve. Any reques
 Throttling is divided into different quota policies:
```
abhilasharora: This could be rephrased as follows or something along those lines to improve clarity.
`The requests to the configuration store are throttled based on the following quota policies.`

docs/REST/throttling.md
Yiming-Jia: Removed `and delete`

docs/REST/throttling.md
abhilasharora: We could use a numbered list, bullet list or a table to organize these four quota policies.

Yiming-Jia: We don't want to list all policies here specifically. There are several short-term throttling policies are not listed. Here we explained quota we calculated for throttling

abhilasharora: Yes, my comment was to make them numbered such as the following, or use bullet points.

1. **TotalRequests** - total amount of requests/sec
1. **WriteRequests** - total amount of write requests/second
1. **Bandwidth** - outbound data in bytes
1. **Storage** - total storage size of user data in bytes

jimmyca15: I don't mind either way. We have a minimilistic approach in the REST docs and to me the new lines are fine without bullets. If you all do decide to do bullets, please don't number them. There is no order associated with these quotas.

abhilasharora: The different quota policies are basically list items, so it makes sense to structure them that way. I'm fine with bullets, similar to how the operations are structured in our other document.

https://github.com/Azure/AppConfiguration/blob/master/docs/REST/kv.md

docs/REST/throttling.md
```diff
@@ -5,9 +5,13 @@ Configuration stores have limits on the requests that they may serve. Any reques
...
 
 **TotalRequests** - total amount of requests/sec
```
zhenlan: NIT: 'sec' -> 'second' to be consistent

docs/REST/throttling.md
```diff
@@ -5,9 +5,13 @@ Configuration stores have limits on the requests that they may serve. Any reques
...
-**TotalWrites** - total amount of write requests/second
+		
+**WriteRequests** - total amount of write requests/second
+
 **Bandwidth** - outbound data in bytes
```
zhenlan: Should this be "outbound data in bytes/second"?

docs/REST/throttling.md
Yiming-Jia: I start a new pr in our repo, will fix those here first: https://msazure.visualstudio.com/Azure%20AppConfig/_git/AppConfigService/pullrequest/2668448?_a=files

docs/REST/throttling.md
jimmyca15: Good point, but I think it's the opposite. The time period should be omitted from these descriptions. Each of these different policies could be throttled over the span of a second/minute/hour/day

@Yiming-Jia

zhenlan: Makes sense. Wording wise, we can say ".... per time unit". But how can a customer figure out it's second, minute or something else?

Yiming-Jia: We can document our throttling policies somewhere in our official document(https://azure.microsoft.com/en-us/services/app-configuration). Here we only explain restful api related concepts.

jimmyca15: @zhenlan from the error object it can't be deduced. @drago-draganov, we may want to surface this information.

Also, I went back and looked and these policy names are outdated. Will work with Yiming on his new PR to make sure that the old names are updated.

PR #307, opened 2020-04-04
Update release notes for Microsoft.Extensions.Configuration.AzureAppConfiguration v3.0.1

abhilasharora:


releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
```diff
@@ -1,6 +1,14 @@
 ## Microsoft.Extensions.Configuration.AzureAppConfiguration
 ### [Package (NuGet)](https://www.nuget.org/packages/Microsoft.Extensions.Configuration.AzureAppConfiguration)
 
+### 3.0.1 - April 02, 2020
```
zhenlan, line 4: For those changes that have corresponding issues, please add references to them (as you did in the previous release) for any who are interested to find out more information.

releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
```diff
@@ -1,6 +1,14 @@
...
+* Improved refresh for feature flags to reduce the number of calls made to App Configuration if no change is detected.
+* Fixed the issue that caused `TryRefreshAsync` to throw an `Azure.Identity.AuthenticationFailedException` when the `TokenCredential` used to fetch a key vault reference failed to authenticate.
+* Fixed the following issues when `optional` parameter is set to `true` in the method `AddAzureAppConfiguration`
+    * An exception might be thrown if the configuration store could not be accessed.
+    * `RefreshAsync` might ignore exceptions if configuration fails to load after a change is detected in a key with `refreshAll: true`.
+    * `TryRefreshAsync` would throw a `NullReferenceException` if the initial attempt to load the configuration in `IConfiguration.Build` failed.
+
```
zhenlan, line 11: Being able to recover from an initial load-failure is a great feature. Please do have a separate bullet point for it and explain how to benefit from it.

releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
```diff
@@ -1,6 +1,28 @@
...
+   IConfigurationRefresher refresher;
+   IConfigurationBuilder configBuilder = new ConfigurationBuilder();
+   IConfiguration configuration = configBuilder.AddAzureAppConfiguration(options => {
+      options.Connect(endpoint, new DefaultAzureCredential());
+      options.ConfigureRefresh(refresh => refresh.Register("some_key"));
+      refresher = options.GetRefresher();
+   }, optional: true);
+   
+   // When the initial configuration fails to load, this attemps to auto-recover the configuration and returns the result
+   bool isRefreshSuccessful = await refresher.TryRefreshAsync();
```
jimmyca15, line 23: I'm wondering if we should document this code snippet? Using it as an optional recovery mechanism is a bit lacking in my opinion unless we expose an API which tells us whether the provider is initialized and, if not, can be recovered. Otherwise I think just documenting that RefreshAsync will now attempt to initialize failed optional loads sounds fine.

jimmyca15: I mean, I wouldn't expect users who aren't already calling RefreshAsync to start calling it because of this new behavior. It's not designed that way. If it was we would have a way to let them know they should call it.

PR #311, opened 2020-04-06
BCDR code snippet: Azure Function polling Azure Queue for events

avanigupta:
Creating a draft PR for review and tracking. Code will not be checked in to the repository.

	1. We set up Event Grid events emitted by Azure App Configuration to be routed to an Azure Storage queue. Note that event order is not guaranteed by Event grid or storage queue.  
	2. We create a timer triggered Azure function to read contents of storage queue every <10> minutes. In this tutorial, we are using the same storage account for this Azure function and Storage Queue. 
	3. Azure function checks if there are any messages from event grid in the queue. Any other messages in the queue in invalid format will be discarded.
	4. If there are any messages in the queue, retrieve a batch of maximum 32 messages at a time for processing.
	5. For all retrieved messages, extract the key+label from event message and store them in a hash set. We need to keep track of any unique key+label that was modified or deleted.
	6. Fetch all settings from primary store. 
	7. Update only those settings in secondary store which have a corresponding event in the storage queue.
	8. Delete all settings that were present in storage queue but not in primary store.
    9. Delete all messages that we received in this batch from the storage queue.
    10. Repeat from step 4 if there are any messages remaining in storage queue.


abhilasharora:
Could you also copy the steps that you included in the OneNote as a comment at the start of the file?

examples/DotNetCore/BackupStore/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,146 @@
...
+            try
+            {
+                log.LogInformation($"C# Timer trigger function executed at: {DateTime.Now}");
+
+                // Connect to Azure App Configuration stores
+                ConfigurationClient primaryAppConfigClient = new ConfigurationClient(new Uri(Environment.GetEnvironmentVariable("PrimaryStoreEndpoint")), new ManagedIdentityCredential());
+                ConfigurationClient secondaryAppConfigClient = new ConfigurationClient(new Uri(Environment.GetEnvironmentVariable("SecondaryStoreEndpoint")), new ManagedIdentityCredential());
+
+                // Connect to storage queue
+                CloudStorageAccount storageAccount = CloudStorageAccount.Parse(Environment.GetEnvironmentVariable("StorageQueueConnectionString"));
```
abhilasharora, line 30: We can re-use the storage account associated with the Azure Function, so the user doesn not need to create a new one or specify a connection string for it.

examples/DotNetCore/BackupStore/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,146 @@
...
+                // return if there are no messages in the queue
+                if (peekedMessage == null)
+                {
+                    log.LogInformation($"No events received for processing.");
+                    return;
+                }
+
+                // Read all settings from Primary Store and create local copy
+                IEnumerable<ConfigurationSetting> primaryStoreSettings = primaryAppConfigClient.GetConfigurationSettings(new SettingSelector());
+                IDictionary<Tuple<string, string>, ConfigurationSetting> primaryStoreLocalSettings = new Dictionary<Tuple<string, string>, ConfigurationSetting>();
```
abhilasharora, line 71: Could we use a custom type instead of a `Tuple` to make it clear what the individual strings are?

examples/DotNetCore/BackupStore/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,146 @@
...
+                foreach (ConfigurationSetting setting in primaryStoreSettings)
+                {
+                    var keyLabel = new Tuple<string, string>(setting.Key, setting.Label);
+                    primaryStoreLocalSettings.Add(keyLabel, setting);
+                }
+
+                // Read all settings from secondary store and create local copy
```
abhilasharora, line 78: Instead of fetching all configuration settings from both primary and secondary stores, how about fetching settings only from primary store and blindly updating the secondary store?

examples/DotNetCore/BackupStore/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,146 @@
...
+                    var messageBody = peekedMessage.AsString;
+                    try
+                    {
+                        JObject dataObject = JObject.Parse(messageBody);
+                        string topic = dataObject.Property("topic")?.Value.ToString();
+                        if (topic != null && topic == Environment.GetEnvironmentVariable("PrimaryStoreResourceId"))
+                        {
+                            // We have events that need to be processed
+                            // Add custom message to the end of queue
+                            queue.AddMessage(customMessage);
```
abhilasharora, line 48: What happens if this call fails due to a network issue?

examples/DotNetCore/BackupStore/BackupFromStorageQueue.cs
abhilasharora: After thinking a bit more on this, it seems this might be better in cases where we process a queue message but don't detect any changes in the settings. In this case, we would only read settings but not write them.

examples/DotNetCore/BackupStore/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,146 @@
...
+        [FunctionName("BackupFromStorageQueue")]
+        public static void Run([TimerTrigger("0 */10 * * * *")]TimerInfo myTimer, ILogger log)
+        {
+            try
+            {
+                log.LogInformation($"C# Timer trigger function executed at: {DateTime.Now}");
+
+                // Connect to Azure App Configuration stores
+                ConfigurationClient primaryAppConfigClient = new ConfigurationClient(new Uri(Environment.GetEnvironmentVariable("PrimaryStoreEndpoint")), new ManagedIdentityCredential());
+                ConfigurationClient secondaryAppConfigClient = new ConfigurationClient(new Uri(Environment.GetEnvironmentVariable("SecondaryStoreEndpoint")), new ManagedIdentityCredential());
```
abhilasharora, line 27: We don't need to create configuration clients if there are no valid messages to be processed.

examples/DotNetCore/BackupStore/BackupFromStorageQueue.cs
avanigupta: Yes. Consider the case when you have 10 update events in the queue, and you add custom message at 11th position in queue. Before you have fetched all settings from primary store, 11th update event was added to the queue on 12th position.
This 11th event will not be deleted from the queue even though we will update it in the secondary store. So the next time Azure function runs, it may have to process just this one event which has actually already been updated in secondary store.

examples/DotNetCore/BackupStore/BackupFromStorageQueue.cs
avanigupta: done

examples/DotNetCore/BackupStore/BackupFromStorageQueue.cs
avanigupta: done

examples/DotNetCore/BackupStore/BackupFromStorageQueue.cs
avanigupta: done

examples/BCDR/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,172 @@
...
+                            updatedKeyLabels.Add(new KeyLabel(key, label));
+                        }
+                    }
+                    catch (JsonReaderException)
+                    {
+                        // If its not a valid JSON, ignore the message 
+                        log.LogInformation($"Queue message in invalid JSON format will be ignored.\nMessage: {decodedMessage}");
+                    }
+                }
+                catch (FormatException)
```
zhenlan, line 64: If this is only for the base64 decoding, it should only wrap that part.

examples/BCDR/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,172 @@
...
+                    // Event grid will encode events in Base64 format before publishing to storage queue
+                    // Decode the message before parsing it
+                    string decodedMessage = Encoding.UTF8.GetString(Convert.FromBase64String(message.MessageText));
+                    try
+                    {
+                        JObject eventObject = JObject.Parse(decodedMessage);
+                        string topic = eventObject.Property("topic")?.Value.ToString();
+                        if (topic != null && topic == Environment.GetEnvironmentVariable("PrimaryStoreResourceId"))
+                        {
+                            JObject dataObject = JObject.Parse(eventObject.Property("data")?.Value.ToString());
```
zhenlan, line 52: If `dataObject` could be null, we should do null-check before the next line.

examples/BCDR/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,172 @@
...
+                log.LogError($"Service request failed from the server with the following exceptions:");
+                foreach (RequestFailedException ex in exception.InnerExceptions)
+                {
+                    log.LogError($"\nStatus: {ex.Status}\nError Message: {ex.Message}");
+                }
+            }
+        }
+    }
+
+    public class KeyLabel
```
zhenlan, line 147: This should live in its own file and it does not need to be `public`.

examples/BCDR/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,172 @@
...
+        public string Label { get; set; }
+
+        public KeyLabel(string key, string label)
+        {
+            Key = key;
+            Label = label;
+        }
+    }
+
+    public class KeyLabelComparer : IEqualityComparer<KeyLabel>
```
zhenlan, line 159: Live in a separate file.

examples/BCDR/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,172 @@
...
+    {
+        public bool Equals(KeyLabel x, KeyLabel y)
+        {
+            return string.Equals(x.Key, y.Key, StringComparison.InvariantCulture) 
+                && string.Equals(x.Label, y.Label, StringComparison.InvariantCulture);
+        }
+
+        public int GetHashCode(KeyLabel obj)
+        {
+            return obj.Key.GetHashCode();
```
zhenlan, line 169: This will cause items with the same key but different labels to be overwritten in the hashset, right?

examples/BCDR/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,172 @@
...
+            {
+                log.LogInformation($"C# Timer trigger function executed at: {DateTime.Now}");
+                
+                // Create storage queue client
+                QueueClient queueClient = new QueueClient(Environment.GetEnvironmentVariable("AzureWebJobsStorage"), "eventgridqueue");
+
+                // Peek to see if there are events in the queue
+                while (queueClient.PeekMessages().Value.Length > 0 )
+                {
+                    Response<QueueMessage[]> messages = queueClient.ReceiveMessages(maxMessages: 32);
```
zhenlan, line 113: Can we use `ReceiveMessagesAsync` instead?

Please define 32 as const.

examples/BCDR/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,172 @@
...
+                    if(updatedKeyLabels.Count > 0)
+                    {
+                        BackupKeyValues(updatedKeyLabels, log);
+                    }
+
+                    // Delete this batch of events from storage queue
+                    foreach (QueueMessage message in messages.Value)
+                    {
+                        log.LogInformation($"Deleting message: {message.MessageId}");
+                        queueClient.DeleteMessage(message.MessageId, message.PopReceipt);
```
zhenlan, line 128: Can we use `DeleteMessageAsync`?

examples/BCDR/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,172 @@
...
+        {
+            try
+            {
+                log.LogInformation($"C# Timer trigger function executed at: {DateTime.Now}");
+                
+                // Create storage queue client
+                QueueClient queueClient = new QueueClient(Environment.GetEnvironmentVariable("AzureWebJobsStorage"), "eventgridqueue");
+
+                // Peek to see if there are events in the queue
+                while (queueClient.PeekMessages().Value.Length > 0 )
```
zhenlan, line 111: Can we use `PeekMessagesAsync`?

examples/BCDR/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,172 @@
...
+                catch (FormatException)
+                {
+                    // If its not in valid base64 format, ignore the message 
+                    log.LogInformation($"Queue message in invalid Base64 format will be ignored.\nMessage: {message.MessageText}");
+                }
+            }
+            return updatedKeyLabels;
+        }
+
+        private static void BackupKeyValues(HashSet<KeyLabel> updatedKeyLabels, ILogger log)
```
zhenlan, line 73: We should make this function async and use async versions of `ConfigurationClient` APIs.

examples/BCDR/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,172 @@
...
+            // Read all settings from Primary Store and update secondary store
+            foreach (ConfigurationSetting setting in primaryAppConfigClient.GetConfigurationSettings(new SettingSelector()))
+            {
+                KeyLabel primaryStoreKeyLabel = new KeyLabel(setting.Key, setting.Label);
+                if(updatedKeyLabels.Contains(primaryStoreKeyLabel))
+                {
+                    // current setting retrieved from primary store needs to be updated in secondary store
+                    secondaryAppConfigClient.SetConfigurationSetting(setting);
+                    log.LogInformation($"Successfully updated key: {setting.Key} label: {setting.Label}");
+                    updatedKeyLabels.Remove(primaryStoreKeyLabel);
```
zhenlan, line 88: One optimization we can add here is that if `updatedKeyLabels` becomes empty at this point, we can `break` the loop.

examples/BCDR/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,172 @@
...
+            // Delete key labels still present in updatedKeyLabels from secondary store
+            foreach (var keyLabel in updatedKeyLabels)
+            {
+                secondaryAppConfigClient.DeleteConfigurationSetting(keyLabel.Key, keyLabel.Label);
+                log.LogInformation($"Successfully deleted key: {keyLabel.Key} label: {keyLabel.Label} from secondary store.");
+            }
+        }
+
+        [FunctionName("BackupFromStorageQueue")]
+        public static void Run([TimerTrigger("0 */10 * * * *")]TimerInfo myTimer, ILogger log)
```
zhenlan, line 101: Just to make the code easier to read, I would move this function to the beginning of the class.

examples/BCDR/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,172 @@
...
+
+            // Delete key labels still present in updatedKeyLabels from secondary store
+            foreach (var keyLabel in updatedKeyLabels)
+            {
+                secondaryAppConfigClient.DeleteConfigurationSetting(keyLabel.Key, keyLabel.Label);
+                log.LogInformation($"Successfully deleted key: {keyLabel.Key} label: {keyLabel.Label} from secondary store.");
+            }
+        }
+
+        [FunctionName("BackupFromStorageQueue")]
```
zhenlan, line 100: The "StorageQueue" is just the implementation detail. The name of the Azure Function should be something like "BackupAppConfigurationStore".

examples/BCDR/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,172 @@
...
+
+                    // Delete this batch of events from storage queue
+                    foreach (QueueMessage message in messages.Value)
+                    {
+                        log.LogInformation($"Deleting message: {message.MessageId}");
+                        queueClient.DeleteMessage(message.MessageId, message.PopReceipt);
+                    }
+                }
+            }
+            catch (RequestFailedException exception)
```
zhenlan, line 132: Should these error handlings be part of `BackupKeyValues`?

examples/BCDR/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,172 @@
...
+                    log.LogInformation($"Queue message in invalid Base64 format will be ignored.\nMessage: {message.MessageText}");
+                }
+            }
+            return updatedKeyLabels;
+        }
+
+        private static void BackupKeyValues(HashSet<KeyLabel> updatedKeyLabels, ILogger log)
+        {
+            // Connect to Azure App Configuration stores
+            ConfigurationClient primaryAppConfigClient = new ConfigurationClient(new Uri(Environment.GetEnvironmentVariable("PrimaryStoreEndpoint")), new ManagedIdentityCredential());
```
zhenlan, line 76: To make it clearer and avoid reading environment variables again and again, I would define `PrimaryStoreEndpoint` (and others) all as const at the beginning of the class.

examples/BCDR/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,172 @@
...
+                try
+                {
+                    // Event grid will encode events in Base64 format before publishing to storage queue
+                    // Decode the message before parsing it
+                    string decodedMessage = Encoding.UTF8.GetString(Convert.FromBase64String(message.MessageText));
+                    try
+                    {
+                        JObject eventObject = JObject.Parse(decodedMessage);
+                        string topic = eventObject.Property("topic")?.Value.ToString();
+                        if (topic != null && topic == Environment.GetEnvironmentVariable("PrimaryStoreResourceId"))
```
zhenlan, line 50: Defining `PrimaryStoreResourceId` is kind of an unnecessary burden to users. I guess if the message has data field with key and label properties, it should be good enough.

examples/BCDR/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,172 @@
...
+using Microsoft.Azure.WebJobs;
+using Microsoft.Extensions.Logging;
+using Azure.Storage.Queues;
+using Azure;
+using Azure.Data.AppConfiguration;
+using Azure.Identity;
+using System.Linq;
+using Newtonsoft.Json.Linq;
+using Newtonsoft.Json;
+using Azure.Storage.Queues.Models;
```
zhenlan, line 30: Sort and delete usings

examples/BCDR/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,172 @@
...
+
+namespace BackupFromStorage
```
zhenlan, line 32: Storage is an implementation detail. Suggest something like "BackupAppConfiguraitonStore". Similarly for the class name.

examples/BCDR/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,172 @@
...
+                    secondaryAppConfigClient.SetConfigurationSetting(setting);
+                    log.LogInformation($"Successfully updated key: {setting.Key} label: {setting.Label}");
+                    updatedKeyLabels.Remove(primaryStoreKeyLabel);
+                }
+            }
+
+            // Delete key labels still present in updatedKeyLabels from secondary store
+            foreach (var keyLabel in updatedKeyLabels)
+            {
+                secondaryAppConfigClient.DeleteConfigurationSetting(keyLabel.Key, keyLabel.Label);
```
zhenlan, line 95: Can you please make sure this will not throw for things like 404?

examples/BCDR/BackupFromStorageQueue.cs
zhenlan: All right. We have to handle the case when the backup failed, we don't delete the messages.

OK, in case the primary is down, the backup will fail. When this happens, we should just abort and the new attempt will be made when the Azure Function is triggered again next time. We can change `BackupKeyValues` to return bool and decide whether we should abort.

examples/BCDR/BackupFromStorageQueue.cs
abhilasharora: This shouldn't cause such key-values to be overwritten, but it will lead to collisions and increase the runtime of operations for the hash table. We should be able to use a hash function that uses both key and label fields to optimize it.

examples/BCDR/BackupFromStorageQueue.cs
```diff
@@ -0,0 +1,172 @@
...
+            Label = label;
+        }
+    }
+
+    public class KeyLabelComparer : IEqualityComparer<KeyLabel>
+    {
+        public bool Equals(KeyLabel x, KeyLabel y)
+        {
+            return string.Equals(x.Key, y.Key, StringComparison.InvariantCulture) 
+                && string.Equals(x.Label, y.Label, StringComparison.InvariantCulture);
```
abhilasharora, line 164: We should use `StringComparison.Ordinal` since `StringComparison.InvariantCulture` uses culture-sensitive sort rules.

avanigupta: done

examples/BCDR/BackupFromStorageQueue.cs
avanigupta: This will throw RequestFailedException for anything other than 200 (delete successful) and 204 (Key not found, nothing to delete). RequestFailedException is being handled appropriately.

examples/BCDR/BackupFromStorageQueue.cs
avanigupta: done

examples/BCDR/BackupFromStorageQueue.cs
avanigupta: done

examples/BCDR/BackupFromStorageQueue.cs
avanigupta: Removed the check for PrimaryStoreResourceId

examples/BCDR/BackupFromStorageQueue.cs
avanigupta: Done. Also made both clients static, and null-checking them when function is triggered.

examples/BCDR/BackupFromStorageQueue.cs
avanigupta: Moved this exception handling to BackupKeyValues, which now returns a boolean to indicate if anything failed. If there was any exception thrown by AppConfig, we log the exception and abort the function. Queue messages will be available for processing again after default timeout.

examples/BCDR/BackupFromStorageQueue.cs
avanigupta: done

examples/BCDR/BackupFromStorageQueue.cs
avanigupta: done

examples/BCDR/BackupFromStorageQueue.cs
avanigupta: done

examples/BCDR/BackupFromStorageQueue.cs
avanigupta: done

examples/BCDR/BackupFromStorageQueue.cs
avanigupta: done

examples/BCDR/BackupFromStorageQueue.cs
avanigupta: done

examples/BCDR/BackupFromStorageQueue.cs
avanigupta: done

examples/BCDR/BackupFromStorageQueue.cs
avanigupta: Using a hash function based on both key and label to reduce collisions.

examples/BCDR/BackupFromStorageQueue.cs
avanigupta: done

examples/BCDR/BackupFromStorageQueue.cs
avanigupta: done

examples/BCDR/BackupFromStorageQueue.cs
avanigupta: done. Also added null check for key (although I don't think key can ever be null).

examples/BCDR/BackupFromStorageQueue.cs
avanigupta: done. Using one try with multiple catch blocks.

examples/BCDR/BackupFromStorageQueue.cs
abhilasharora: If this throws a `RequestFailedException` with `404`, then the current implementation would get stuck in an infinite processing cycle since subsequent attempts to process these messages would also get the same error. Instead, we can simply ignore such an error here and proceed as normal, since the key-value that we're trying to delete is already in the desired state (non-existent).

examples/ConfigurationStoreBackup/KeyLabelComparer.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace ConfigurationStoreBackup
+{
+    internal class KeyLabelComparer : IEqualityComparer<KeyLabel>
```
abhilasharora, line 6: Instead of using a custom comparator, it might be simpler to move the `Equals` and `GetHashCode` methods within the `KeyLabel` type, so related logic is in the same file.

examples/ConfigurationStoreBackup/KeyLabel.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace ConfigurationStoreBackup
+{
+    internal class KeyLabel
```
abhilasharora, line 3: This could be a `struct` instead of a `class` so we could get value semantics.

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,177 @@
...
+using System.Threading.Tasks;
+
+namespace ConfigurationStoreBackup
+{
+    public static class BackupAppConfigurationStore
+    {
+        private static readonly ConfigurationClient primaryAppConfigClient = new ConfigurationClient(new Uri(Environment.GetEnvironmentVariable("PrimaryStoreEndpoint")), new ManagedIdentityCredential());
+        private static readonly ConfigurationClient secondaryAppConfigClient = new ConfigurationClient(new Uri(Environment.GetEnvironmentVariable("SecondaryStoreEndpoint")), new ManagedIdentityCredential());
+        private static readonly QueueClient queueClient = new QueueClient(Environment.GetEnvironmentVariable("AzureWebJobsStorage"), "eventgridqueue");
+        private static readonly int maxMessagesToRead = 32;
```
abhilasharora, line 40: This could be a `const` value.

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,177 @@
...
+                // Delete this batch of events from storage queue.
+                foreach (QueueMessage message in retrievedMessages.Value)
+                {
+                    log.LogDebug($"Deleting message: {message.MessageId}");
+                    await queueClient.DeleteMessageAsync(message.MessageId, message.PopReceipt).ConfigureAwait(false);
+                }
+            }
+        }
+
+        private static async Task<bool> AnyMessageInQueueAsync()
```
abhilasharora, line 88: The method name seems a bit odd. Could the check be done inline?

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,177 @@
...
+                {
+                    // If its not in valid base64 format, ignore the queue message. 
+                    log.LogInformation($"Queue message in invalid Base64 format will be ignored.\nMessage: {message.MessageText}");
+                }
+                catch (JsonReaderException)
+                {
+                    // If its not a valid JSON, ignore the queue message. 
+                    log.LogInformation($"Queue message in invalid JSON format will be ignored.\nMessage: {decodedMessage}");
+                }
+                catch (NullReferenceException)
```
abhilasharora, line 122: We shouldn't catch `NullReferenceException` since it doesn't necessarily mean it was thrown due to `key` being `null`. We can add an `if` check for `key` and ignore message if it is `null`.

examples/BCDR/BackupFromStorageQueue.cs
avanigupta: If the key value is already in desired state (non-existent), it will not throw 404 with exception. Instead, it returns 204, which is considered a success response. If we still get 404 somehow, it means something is actually wrong, and we abort backing up any key values. In fact, if there is any exception thrown by our SDK, we abort the current function execution and rely on the next triggered execution to process the entire backlog of event messages.

abhilasharora: Okay, I just checked the SDK code and it does not seem to throw `404`. It dies however throw `409` if the setting is read-only. Have you tried replicating a read-only setting?

avanigupta: Yes, if users have somehow locked a setting in their secondary store and update/delete the same setting in primary store, we will keep hitting this exception and crashing the function on every run. Since we currently do not support backing up locking/unlocking of keys, the only way to hit this exception is if users manually lock a key in secondary store.

abhilasharora: Here's a scenario. A user edits a setting and locks the key. When we receive the event for setting being updated, we pull the read only setting and write to the secondary store. Now if the user unlocks the setting and updates it, we will hit this issue and the backup would get stuck.

avanigupta: Looking at the SDK code, we dont actually set key as readonly with the SetConfigurationSettingAsync(ConfigurationSetting) call. There is a separate function for creating/updating read only keys called SetReadOnlyAsync (https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/appconfiguration/Azure.Data.AppConfiguration/src/ConfigurationClient.cs#L887 ).

We can handle this in two ways:
1. We don't bother locking keys in secondary store at all.
2. We check if a setting is readonly, and call the appropriate SetReadOnlyAsync API for locked keys.

The problem with second approach is - we don't receive any events if key is simply locked/unlocked in primary store. So the only way to backup lock/unlock changes is if they are accompanied with update/delete changes. In my opinion, it would just create an unpredictable inconsistency between the primary and secondary stores (and also make the sample code snippet more complex).

abhilasharora: So in the current implementation, we do support backup for read-only keys, but the read-only flag would not be persisted in the secondary store. This should be fine for now. When we do start firing events for lock/unlock changes, we can revisit this and see if we can also backup the read-only flag.

avanigupta: Sounds good.

zhenlan: Let's make it explicit. Avani, in the instructions, please call out subscribing only to the "modified" and "deleted" events, so we don't need to worry about whether/when we add the "lock-modified" event in the future.

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,166 @@
...
+                catch (FormatException)
+                {
+                    // If its not in valid base64 format, ignore the queue message. 
+                    log.LogInformation($"Queue message in invalid Base64 format will be ignored.\nMessage: {message.MessageText}");
+                }
+                catch (JsonReaderException)
+                {
+                    // If its not a valid JSON, ignore the queue message. 
+                    log.LogInformation($"Queue message in invalid JSON format will be ignored.\nMessage: {decodedMessage}");
+                }
```
abhilasharora, line 115: We can skip these comments since the log statements already capture this information. Also re-evaluate other comments that are followed by log statements.

examples/ConfigurationStoreBackup/KeyLabel.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+
+namespace ConfigurationStoreBackup
+{
+    internal struct KeyLabel
+    {
+        internal string Key { get; }
+        internal string Label { get; }
```
abhilasharora, line 8: The could be public since these are part of an internal class.

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,166 @@
...
+                    if (updatedKeyLabels.Contains(primaryStoreKeyLabel))
+                    {
+                        // Current setting retrieved from primary store needs to be updated in secondary store.
+                        await secondaryAppConfigClient.SetConfigurationSettingAsync(setting).ConfigureAwait(false);
+                        log.LogInformation($"Successfully updated key: {setting.Key} label: {setting.Label}");
+                        
+                        updatedKeyLabels.Remove(primaryStoreKeyLabel);
+                        if (updatedKeyLabels.Count == 0)
+                        {
+                            // If no more key labels were updated, we can stop processing.
```
abhilasharora, line 138: > If no more key labels were updated, we can stop processing. [](start = 30, length = 60)

This condition is already true inside the if block. Not sure if this comment is needed.

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,166 @@
...
+                    // If its not a valid JSON, ignore the queue message. 
+                    log.LogInformation($"Queue message in invalid JSON format will be ignored.\nMessage: {decodedMessage}");
+                }
+            }
+            return updatedKeyLabels;
+        }
+
+        private static async Task<bool> BackupKeyValuesAsync(HashSet<KeyLabel> updatedKeyLabels, ILogger log)
+        {
+            bool backupCompleted = false;
```
abhilasharora, line 122: >backupCompleted [](start = 17, length = 15)

How about `isBackupComplete` or `isBackupSuccessful`?

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,163 @@
+/*
```
abhilasharora, line 1: If you're using an Azure Functions app that you're using to test this code, we might want to include the project file here. Then users can just download the project, update the connection strings and trigger the function. This will make things easier since they don't have to create their own project (in case they don't have one) and add Nuget package references.

avanigupta: That's a good idea. Added the project and solution files.

examples/ConfigurationStoreBackup/ConfigurationStoreBackup.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <AzureFunctionsVersion>v3</AzureFunctionsVersion>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Azure.Data.AppConfiguration" Version="1.0.0" />
+    <PackageReference Include="Azure.Identity" Version="1.1.1" />
+    <PackageReference Include="Azure.Storage.Queues" Version="12.0.0" />
+    <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="3.0.3" />
```
abhilasharora, line 10: We can use the latest patched version `3.0.6` for this package.

examples/ConfigurationStoreBackup/host.json
```diff
@@ -0,0 +1,11 @@
+{
+    "version": "2.0",
+    "logging": {
```
zhenlan, line 3: I think we can remove the whole "logging" section. We don't need AppInsights.

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,163 @@
...
+        private static readonly ConfigurationClient secondaryAppConfigClient = new ConfigurationClient(new Uri(Environment.GetEnvironmentVariable("SecondaryStoreEndpoint")), new ManagedIdentityCredential());
+        private static readonly QueueClient queueClient = new QueueClient(Environment.GetEnvironmentVariable("AzureWebJobsStorage"), "eventgridqueue");
+        private const int maxMessagesToRead = 32;
+
+        [FunctionName("BackupAppConfigurationStore")]
+        public static void Run([TimerTrigger("0 */10 * * * *")]TimerInfo myTimer, ILogger log)
+        {
+            log.LogInformation($"C# Timer trigger function executed at: {DateTime.Now}");
+
+            if (primaryAppConfigClient == null || secondaryAppConfigClient == null)
```
zhenlan, line 47: Shouldn't we validate `queueClient` too?

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,163 @@
...
+            {
+                log.LogError($"Could not connect to App Configuration Store. Please validate the store endpoints that you have provided.");
+                return;
+            }
+
+            // Block the thread so that another function instance cannot be triggered while this is still processing.
+            ProcessStorageQueueMessagesAsync(log).ConfigureAwait(false).GetAwaiter().GetResult();
```
zhenlan, line 54: We are not writing libraries. We don't need to call `ConfigureAwait(false)` (here and the rest of the code).

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,163 @@
...
+                    if (dataObject != null && dataObject["key"] != null)
+                    {
+                        string key = dataObject["key"].ToString();
+                        string label = dataObject["label"]?.ToString();
+                        updatedKeyLabels.Add(new KeyLabel(key, label));
+                    }
+                }
+                catch (FormatException)
+                {
+                    log.LogInformation($"Queue message in invalid Base64 format will be ignored.\nMessage: {message.MessageText}");
```
zhenlan, line 108: When someone see this log message, the message is already ignored :). Suggest to change "will be" to "is". Same for the log message below.

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,163 @@
...
+                // Delete key labels still present in updatedKeyLabels from secondary store.
+                foreach (var keyLabel in updatedKeyLabels)
+                {
+                    await secondaryAppConfigClient.DeleteConfigurationSettingAsync(keyLabel.Key, keyLabel.Label).ConfigureAwait(false);
+                }
+                isBackupSuccessful = true;
+            }
+            catch (RequestFailedException exception)
+            {
+                log.LogError($"Service request failed from the server.\nStatus: {exception.Status}\nError Message: {exception.Message}");
```
zhenlan, line 150: I hope we can have clearer error message here. How about "Request to Azure App Configuration failed...."

Maybe we should include call stack. It's hard to tell whether it's request to the primary or the secondary failed. Is there an easy way?

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,163 @@
...
+            }
+            catch (AggregateException exception) when ((exception as AggregateException)?.InnerExceptions?.All(ex => ex is RequestFailedException) ?? false)
+            {
+                log.LogError($"Service request failed from the server with the following exceptions:");
```
zhenlan, line 154: "Request to Azure App Configuration failed...."

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,163 @@
...
+                HashSet<KeyLabel> updatedKeyLabels = ExtractKeyLabelsFromEvents(retrievedMessages.Value, log);
+
+                // If there are any valid App Configuration events, update secondary store.
+                if (updatedKeyLabels.Count > 0)
+                {
+                    bool isBackupSuccessful = await BackupKeyValuesAsync(updatedKeyLabels, log).ConfigureAwait(false);
+                    if (!isBackupSuccessful)
+                    {
+                        // Abort this function without deleting retrievedMessages from storage queue.
+                        log.LogError($"Backup failed because of error(s) returned by Azure App Configuration service.");
```
zhenlan, line 74: Suggest to change this to "App Configuration store backup is aborted. It will be attempted next time."

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,163 @@
...
+                        return;
```
zhenlan, line 75: Will `break;` be better option here? It avoids terminating from the middle of a function.

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,163 @@
...
+            // Block the thread so that another function instance cannot be triggered while this is still processing.
+            ProcessStorageQueueMessagesAsync(log).ConfigureAwait(false).GetAwaiter().GetResult();
+        }
+
+        private static async Task ProcessStorageQueueMessagesAsync(ILogger log)
+        {
+            // Peek to see if there are events in the queue.
+            while ((await queueClient.PeekMessagesAsync().ConfigureAwait(false)).Value.Length > 0)
+            {
+                Response<QueueMessage[]> retrievedMessages = await queueClient.ReceiveMessagesAsync(maxMessagesToRead).ConfigureAwait(false);
```
zhenlan, line 62: What transient error this call can throw? We should swallow those errors that can be ignored. We certainly don't want the Function crash for those.

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,163 @@
...
+                        log.LogError($"Backup failed because of error(s) returned by Azure App Configuration service.");
+                        return;
+                    }
+                }
+
+                // Delete this batch of events from storage queue.
+                foreach (QueueMessage message in retrievedMessages.Value)
+                {
+                    log.LogDebug($"Deleting message: {message.MessageId}");
+                    await queueClient.DeleteMessageAsync(message.MessageId, message.PopReceipt).ConfigureAwait(false);
```
zhenlan, line 83: Any transient error this may throw? We shouldn't crash the Function app for those.

zhenlan:
@avanigupta This looks good. After the code is finalized, we should set this up and keep it running for a few days to make sure there are no unexpected errors, etc.

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
avanigupta: I looked at all the properties of this exception. There is no way to tell which config store is throwing the exception. We can add call stack as it can help isolate the call that is throwing the exception.
This is all the data we get in the exception for reference:

Status: 409
Error Message: The setting is read only
 Status: 409 (Conflict)

 Content:
 {"type":"[https://azconfig.io/errors/key-locked](https://azconfig.io/errors/key-locked)","title":"Modifing key 'abc' is not allowed","name":"abc","detail":"The key is read-only. To allow modification unlock it first.","status":409}

 Headers:
 Server: nginx/1.16.1
 Date: Wed, 22 Apr 2020 18:50:11 GMT
 Transfer-Encoding: chunked
 Connection: keep-alive
 x-ms-request-id: 9354d085-cf8c-4486-b8c5-b09b66ba6add
 x-ms-client-request-id: d5baa60f-9008-4c44-ad99-344ad459c352
 x-ms-correlation-request-id: 9354d085-cf8c-4486-b8c5-b09b66ba6add
 Access-Control-Allow-Origin: *
 Access-Control-Allow-Credentials: true
 Access-Control-Expose-Headers: DNT, X-CustomHeader, Keep-Alive, User-Agent, X-Requested-With, If-Modified-Since, Cache-Control, Content-Type, Authorization, x-ms-client-request-id, x-ms-useragent, x-ms-content-sha256, x-ms-date, host, Accept, Accept-Datetime, Date, If-Match, If-None-Match, Sync-Token, x-ms-return-client-request-id, ETag, Last-Modified, Link, Memento-Datetime, x-ms-retry-after, x-ms-request-id, WWW-Authenticate
 Strict-Transport-Security: max-age=15724800; includeSubDomains
 Content-Type: application/problem+json; charset=utf-8

Stack Trace:    at Azure.Data.AppConfiguration.ConfigurationClient.DeleteConfigurationSettingAsync(String key, String label, MatchConditions requestOptions, CancellationToken cancellationToken)
   at Azure.Data.AppConfiguration.ConfigurationClient.DeleteConfigurationSettingAsync(String key, String label, CancellationToken cancellationToken)
   at ConfigurationStoreBackup.BackupAppConfigurationStore.BackupKeyValuesAsync(HashSet`1 updatedKeyLabels, ILogger log) in D:\SampleProjects\bcdr\BackupFromStorage\BackupFromStorage\BackupAppConfigurationStore.cs:line 149

Data: System.Collections.ListDictionaryInternal
ErrorCode : 
HelpLink : 
HResult : -2146233088
InnerException :
Source: Azure.Data.AppConfiguration
TargetSite: Void MoveNext()



---
In reply to: [412703340](https://github.com/Azure/AppConfiguration/pull/311#discussion_r412703340) [](ancestors = 412703340)

zhenlan: I would just dump out this whole thing. The request-id is in it. At least, customers can contact us if they think this is unexpected.

avanigupta: Request ID is already included in exception.Message. 

---
In reply to: [413289512](https://github.com/Azure/AppConfiguration/pull/311#discussion_r413289512) [](ancestors = 413289512)

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
avanigupta: Reading messages (get and peek both) will not throw any transient errors. 

---
In reply to: [412706456](https://github.com/Azure/AppConfiguration/pull/311#discussion_r412706456) [](ancestors = 412706456)

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
avanigupta: This may throw transient errors like - Pop receipt mismatch, or message not found. We can ignore those and continue processing.

---
In reply to: [412707275](https://github.com/Azure/AppConfiguration/pull/311#discussion_r412707275) [](ancestors = 412707275)

examples/ConfigurationStoreBackup/host.json
avanigupta: done

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
avanigupta: done

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
avanigupta: done

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
avanigupta: done

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
avanigupta: done

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
avanigupta: done

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
avanigupta: done

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,179 @@
...
+                return;
+            }
+            if (queueClient == null)
+            {
+                log.LogError($"Could not connect to storage queue. Please validate the storage connection string and storage queue name.");
+                return;
+            }
+
+            // Block the thread so that another function instance cannot be triggered while this is still processing.
+            ProcessStorageQueueMessagesAsync(log).GetAwaiter().GetResult();
```
zhenlan, line 59: I suppose you can just do
```cs
ProcessStorageQueueMessagesAsync(log).Wait();
```

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,179 @@
...
+{
+    public static class BackupAppConfigurationStore
+    {
+        private static readonly ConfigurationClient primaryAppConfigClient = new ConfigurationClient(new Uri(Environment.GetEnvironmentVariable("PrimaryStoreEndpoint")), new ManagedIdentityCredential());
+        private static readonly ConfigurationClient secondaryAppConfigClient = new ConfigurationClient(new Uri(Environment.GetEnvironmentVariable("SecondaryStoreEndpoint")), new ManagedIdentityCredential());
+        private static readonly QueueClient queueClient = new QueueClient(Environment.GetEnvironmentVariable("AzureWebJobsStorage"), "eventgridqueue");
+        private const int maxMessagesToRead = 32;
+
+        [FunctionName("BackupAppConfigurationStore")]
+        public static void Run([TimerTrigger("0 */10 * * * *")]TimerInfo myTimer, ILogger log)
```
zhenlan, line 43: I'm wondering whether it will work if we change this to async. Can you give it a try since you are on it?
```cs
public static async Task Run(...)
```

abhilasharora: Yes, this should work. Azure functions could be marked async.

abhilasharora: I wonder though if the function was made synchronous on purpose since I see this comment in this method implementation.

`Block the thread so that another function instance cannot be triggered while this is still processing.`

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,179 @@
...
+            {
+                log.LogError($"Could not connect to storage queue. Please validate the storage connection string and storage queue name.");
+                return;
+            }
+
+            // Block the thread so that another function instance cannot be triggered while this is still processing.
+            ProcessStorageQueueMessagesAsync(log).GetAwaiter().GetResult();
+        }
+
+        private static async Task ProcessStorageQueueMessagesAsync(ILogger log)
```
zhenlan, line 62: It's not all about processing messages. How about `BackupAppConfigurationStoreAsync(...)'?

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,179 @@
...
+            log.LogInformation($"C# Timer trigger function executed at: {DateTime.Now}");
+
+            if (primaryAppConfigClient == null || secondaryAppConfigClient == null)
+            {
+                log.LogError($"Could not connect to App Configuration Store. Please validate the store endpoints that you have provided.");
+                return;
+            }
+            if (queueClient == null)
+            {
+                log.LogError($"Could not connect to storage queue. Please validate the storage connection string and storage queue name.");
```
zhenlan, line 54: How about $"Could not connect to storage queue. Please make sure the environment variable 'AzureWebJobsStorage` is set to the storage connection string and storage queue name.".

I actually don't know what the environment variable needs to be set to. Please correct the text accordingly.

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,179 @@
...
+        private static readonly ConfigurationClient secondaryAppConfigClient = new ConfigurationClient(new Uri(Environment.GetEnvironmentVariable("SecondaryStoreEndpoint")), new ManagedIdentityCredential());
+        private static readonly QueueClient queueClient = new QueueClient(Environment.GetEnvironmentVariable("AzureWebJobsStorage"), "eventgridqueue");
+        private const int maxMessagesToRead = 32;
+
+        [FunctionName("BackupAppConfigurationStore")]
+        public static void Run([TimerTrigger("0 */10 * * * *")]TimerInfo myTimer, ILogger log)
+        {
+            log.LogInformation($"C# Timer trigger function executed at: {DateTime.Now}");
+
+            if (primaryAppConfigClient == null || secondaryAppConfigClient == null)
```
zhenlan, line 47: I would separate the validation of these so we can give clear message which one is wrong.

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
abhilasharora: We could shorten it to `"Could not connect to storage queue. Please make sure the environment variable 'AzureWebJobsStorage' is correctly set"` since this value could be `UseDevelopmentStorage=true` when the user uses a storage emulator for local testing.
 
This is set in the settings file in the function app (which we can add to our example code) or in the function app settings in Azure Portal.

examples/BCDR/BackupFromStorageQueue.cs
avanigupta: Added explicit instructions in the tutorial to subscribe to only KV modified and deleted events.

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
avanigupta: Yes, we can make the function async, but we want only one function instance to process events at a time, which is the intended behavior of timer triggered functions.

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
avanigupta: Added the app settings file for reference and made the error message generic like Abhilash suggested. The documentation also has instructions on setting this to storage account connection string.

examples/ConfigurationStoreBackup/local.settings.json
```diff
@@ -0,0 +1,9 @@
+{
```
zhenlan, line 1: This file is for 'local' use. It should not be checked in.

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,216 @@
...
+using System.Threading.Tasks;
+
+namespace ConfigurationStoreBackup
+{
+    public static class BackupAppConfigurationStore
+    {
+        
+        private static readonly string storageConnectionString = Environment.GetEnvironmentVariable("AzureWebJobsStorage");
+        private static readonly string primaryStoreEndpoint = Environment.GetEnvironmentVariable("PrimaryStoreEndpoint");
+        private static readonly string secondaryStoreEndpoint = Environment.GetEnvironmentVariable("SecondaryStoreEndpoint");
```
zhenlan, line 40: This is what I would suggest

```cs
private const string PrimaryConfigStoreEndpointEnvVarName   = "PrimaryConfigStoreEndpoint"; // eg., https://{store1}.azcofig.io
private const string SecondaryConfigStoreEndpointEnvVarName = "SecondaryConfigStoreEndpoint"; // eg., https://{store2}.azcofig.io
private const string StorageQueueUriEnvVarName = "StorageQueueUri"; // eg., https://{account_name}.queue.core.windows.net/{queue_name}
...

public static void Run(...)
{
    ...
    string primaryStoreEndpoint = Environment.GetEnvironmentVariable(PrimaryConfigStoreEndpointEnvVarName);
    if (string.IsNullOrEmpty(primaryStoreEndpoint))
    {
        log.LogError($"Please ensure that the environment variable '{PrimaryConfigStoreEndpointEnvVarName}' is set to the endpoint of the primary App Configuration store.");
        return;
    }
    ...
}
```

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,216 @@
...
+            }
+            if (string.IsNullOrEmpty(secondaryStoreEndpoint))
+            {
+                log.LogError($"Please ensure that the environment variable 'SecondaryStoreEndpoint' is set to your secondary store endpoint.");
+                return;
+            }
+
+            try
+            {
+                QueueClient queueClient = new QueueClient(storageConnectionString, QueueName);
```
zhenlan, line 66: Looks like the storage queue supports `TokenCredential` too [`QueueClient(Uri, TokenCredential, QueueClientOptions)`](https://docs.microsoft.com/en-us/dotnet/api/azure.storage.queues.queueclient.-ctor?view=azure-dotnet#Azure_Storage_Queues_QueueClient__ctor_System_Uri_Azure_Core_TokenCredential_Azure_Storage_Queues_QueueClientOptions_). We should use managed identity for the storage queue as well.

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
zhenlan: Let's say a backup operation takes 30s and this Function is triggered every 10s. Even in the sync version of the Function, what happens? Will the Function not be triggered until the previous one is completed? 

In the same scenario but an async version of the Function, we will `await` on the `BackupAppConfigurationStoreAsync`. Will the Function be triggered with overlaps?

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,216 @@
...
+                    catch (RequestFailedException ex) when (
+                        ex.ErrorCode == QueueErrorCode.PopReceiptMismatch ||
+                        ex.ErrorCode == QueueErrorCode.MessageNotFound)
+                    {
+                        // We can continue processing the rest of the queue and safely ignore these exceptions.
+                        log.LogDebug($"Failed to delete message from queue. Message ID: {message.MessageId}\nException: {ex}");
+                    }
+                }
+            }
+
```
zhenlan, line 133: NIT: remove empty line

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,216 @@
...
+                log.LogError($"Failed to connect to App Configuration store with Managed Identity credentials.\nException: {ex.Message}");
+            }
+            catch (RequestFailedException ex) when (ex.ErrorCode == QueueErrorCode.QueueNotFound)
+            {
+                log.LogError($"Queue '{QueueName}' does not exist.\nException: {ex.Message}");
+            }
+            catch (RequestFailedException ex)
+            {
+                log.LogError($"Aborting App Configuration store backup.\nService request failed with the following exception: {ex}");
+            }
```
zhenlan, line 88: None of these exceptions are transient errors. I'm leaning towards letting them bubble up and crash the app. Thoughts?

abhilasharora: I agree that we don't need to catch these exceptions. What about transient exceptions that cause the backup to fail? Should we also let these terminate the function abnormally, so the Azure Function shows the status of that instance as Failed in the Azure Portal?

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
abhilasharora: By design, only one instance of a timer-triggered Azure Function can be active at a time. For the sync version, the subsequent instances of a function will be skipped if a previous instance is running. I would expect the same behavior for async version too, since the function would still be technically running when the next instance is triggered, but we might want to confirm it.

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
avanigupta: I am catching these exceptions in order to provide an actionable error message to the users in these cases. 
The only way to mark the function as "Failed" is to rethrow any exceptions that we catch after logging error message. In our code snippet, the only transient exceptions that we can ignore are coming from parsing queue message and deleting queue messages. 
Everything else coming from AppConfig is non-transient and we abort backup anyway. How about rethrowing those exceptions then?

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
avanigupta: Confirmed using a test function that async version of this function will also be triggered only one instance at a time.

zhenlan: Great. Thanks for the confirmation, @avanigupta. In this case, we can have a more natural async flow code.

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
zhenlan: What will users get when the Function crashes? I suppose there are logs?

avanigupta: Yes, there will be logs either way. If the function crashes, users have the option of acting on that failure using AppInsights or LogAnalytics.

zhenlan: That sounds good enough for users to debug their code.

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
avanigupta: As discussed offline, we can continue using connection string for this code because this environment variable exists by default for every Function App. Since the documentation will explicitly ask users to use the same storage account for Function App and Storage Queue, we can reuse the connection string for authentication here.

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,184 @@
...
+                foreach (var keyLabel in updatedKeyLabels)
+                {
+                    await secondaryAppConfigClient.DeleteConfigurationSettingAsync(keyLabel.Key, keyLabel.Label);
+                    log.LogInformation($"Successfully deleted key: {keyLabel.Key} label: {keyLabel.Label}");
+                }
+            }
+            catch (RequestFailedException)
+            {
+                log.LogError($"Request to Azure App Configuration failed.");
+                throw;
```
zhenlan, line 175: Why we change this? I thought we didn't want to crash the Function for transient errors.

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,184 @@
...
+        private const int MaxMessagesToRead = 32;
+
+        [FunctionName("BackupAppConfigurationStore")]
+        public static async Task Run([TimerTrigger("0 */10 * * * *")]TimerInfo myTimer, ILogger log)
+        {
+            log.LogInformation($"Azure App Configuration store backup started at: {DateTime.Now}");
+            string storageConnectionString = Environment.GetEnvironmentVariable(StorageConnectionStringEnvVarName);
+            if (string.IsNullOrEmpty(storageConnectionString))
+            {
+               throw new InvalidOperationException($"Please ensure that the environment variable '{StorageConnectionStringEnvVarName}' is correctly set.");
```
zhenlan, line 50: `ArgumentException` probably will be a better choice.

Also, can we be clear about what the value needs to be set to, for example "... the environment variable '{StorageConnectionStringEnvVarName}' is set to the connection string of the storage account"?

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,184 @@
...
+                        updatedKeyLabels.Add(new KeyLabel(key, label));
+                    }
+                }
+                catch (FormatException)
+                {
+                    log.LogInformation($"Queue message in invalid Base64 format is ignored.\nMessage: {message.MessageText}");
+                }
+                catch (JsonReaderException)
+                {
+                    log.LogInformation($"Queue message in invalid JSON format is ignored.\nMessage: {decodedMessage}");
```
zhenlan, line 134: Should this be `LogError`?

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,184 @@
...
+                    await BackupKeyValuesAsync(updatedKeyLabels, primaryAppConfigClient, secondaryAppConfigClient, log);
+                }
+
+                // Delete this batch of events from storage queue.
+                foreach (QueueMessage message in retrievedMessages.Value)
+                {
+                    try
+                    {
+                        await queueClient.DeleteMessageAsync(message.MessageId, message.PopReceipt);
+                        log.LogDebug($"Sucessfully deleted message from queue. Message ID: {message.MessageId}");
```
zhenlan, line 97: Should this be `LogInformation`? Can you please make sure we are consistent in other places as well?

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
zhenlan: Yes, don't want to block on this. Let's take this as a separate work item after this code is merged.

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
zhenlan: All right. I realized I was not clear in my previous answer. I was only commenting on the exceptions here.

We need to have a common understanding of transient errors first. Errors we talk about here, for example, the URI does not have a correct format or storage queue cannot be found, are *not* transient because they will not fix themselves no matter how many times you try.

However, during the backup, you may occasionally fail to read queue messages, load config, or set config. If you try again, they may just work fine. These errors are considered transient because you can recover from them. We shouldn't let these errors crash the Function. The backup is aborted but the Function didn't fail. Most of the time, no attention needs to be taken either.

I hope this clarifies a little.

avanigupta: Okay, I think I misunderstood earlier. I thought that any exceptions thrown by App Config are not transient because we are failing to backup config store and that's something users would like to be notified of. But actually the errors from App Config are transient because we may succeed in modifying config on a retry, and no action is needed from users.
The only exception I can think of where users need to take action is when a key being modified is locked in secondary store. In that case, backup will be stuck in an infinite loop unless someone unlocks the key from secondary store. Maybe we can catch that one exception and rethrow?

examples/ConfigurationStoreBackup/BackupAppConfigurationStore.cs
```diff
@@ -0,0 +1,196 @@
...
+
+            string secondaryStoreEndpoint = Environment.GetEnvironmentVariable(SecondaryConfigStoreEndpointEnvVarName);
+            if (string.IsNullOrEmpty(secondaryStoreEndpoint))
+            {
+                throw new ArgumentException($"Please ensure that the environment variable '{SecondaryConfigStoreEndpointEnvVarName}' is set to the endpoint of the secondary App Configuration store.");
+            }
+
+            QueueClient queueClient = new QueueClient(new Uri(storageQueueUri), new ManagedIdentityCredential());
+            ConfigurationClient primaryAppConfigClient = new ConfigurationClient(new Uri(primaryStoreEndpoint), new ManagedIdentityCredential());
+            ConfigurationClient secondaryAppConfigClient = new ConfigurationClient(new Uri(secondaryStoreEndpoint), new ManagedIdentityCredential());
```
zhenlan, line 67: Just realized we didn't have to create instantiate *AppConfigClient here. We only need to create them when there are actually messages in the queue. Can you please move them to `BackupKeyValuesAsync`?

avanigupta: Okay, moved client instantiations to BackupAppConfigurationStoreAsync. Also moved QueueClient because it creates a logical grouping of all client instantiations in one method without any side effects.

examples/ConfigurationStoreBackup/KeyLabel.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public override bool Equals(Object obj)
+        {
+            if (obj is KeyLabel)
+            {
+                KeyLabel keyLabel = (KeyLabel)obj;
+                return string.Equals(this.Key, keyLabel.Key, StringComparison.Ordinal)
+                    && string.Equals(this.Label, keyLabel.Label, StringComparison.Ordinal);
+            }
+            else
```
abhilasharora, line 24: The `else` statement here is redundant.

examples/ConfigurationStoreBackup/KeyLabel.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            Key = key;
+            Label = label;
+        }
+
+        public override bool Equals(Object obj)
+        {
+            if (obj is KeyLabel)
+            {
+                KeyLabel keyLabel = (KeyLabel)obj;
```
abhilasharora, line 20: C# 7.0 and later support cast using pattern matching. So you could simplify this as `if (obj is KeyLabel keyLabel)`

PR #312, opened 2020-04-09
Add EventHub sample code

gsgvijay:


MSGaryWang:
Please remove those generated files

.gitignore
```diff
@@ -331,3 +331,5 @@ ASALocalRun/
 
 # MFractors (Xamarin productivity tool) working folder 
 .mfractor/
+
+appSettings.json
```
jimmyca15, line 335: Why do we want to ignore this? I normally see it included in checkins.

examples/DotNetCore/EventHub/Controllers/HomeController.cs
```diff
@@ -0,0 +1,169 @@
...
+using Azure.Messaging.EventHubs.Consumer;
+using Azure.Messaging.EventHubs.Processor;
+using Newtonsoft.Json;
+using Azure.Messaging.EventHubs;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Configuration.AzureAppConfiguration;
+
+namespace SimpleAppConfigEventHub.Controllers
+{
+    public class HomeController : Controller
```
jimmyca15, line 19: >HomeController [](start = 17, length = 14)

There's a lot going on in this controller. How can we make it so that it knows nothing about event hub?

Can we come up with a service that exposes events that the home controller would care about. That service can abstract event hub

examples/DotNetCore/EventHub/Controllers/HomeController.cs
```diff
@@ -0,0 +1,169 @@
...
+            var storageClient = new BlobContainerClient(connectionString: blobStorageConnectionString,
+                                                        blobContainerName: blobContainerName);
+            
+            _processorClient = new EventProcessorClient(checkpointStore: storageClient,
+                                                        consumerGroup: consumerGroup,
+                                                        connectionString: eventHubConnectionString,
+                                                        eventHubName: eventHubName);
+        }
+
+        private string ReadFromConfigStore(string key, string label)
```
jimmyca15, line 109: >ReadFromConfigStore [](start = 23, length = 19)

This pattern doesn't look like intended usage of our configuration provider. If we need on demand lookups we should probably use the SDK directly. Then we can have a single client.

examples/DotNetCore/EventHub/EventContract.cs
```diff
@@ -0,0 +1,44 @@
+using System;
+using Newtonsoft.Json;
+
+namespace SimpleAppConfigEventHub
+{
+    class EventContract
```
jimmyca15, line 6: >EventContract [](start = 10, length = 13)

This should be KeyValueEvent or ConfigurationSettingEvent. something relating it to Azure App Configuration.

examples/DotNetCore/EventHub/Controllers/HomeController.cs
```diff
@@ -0,0 +1,169 @@
...
+
+            foreach (var e in events)
+            {
+                var key = e.Data.Key;
+                var label = e.Data.Label;
+
+                if (string.Equals(key, "Messages", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    var value = ReadFromConfigStore(key, label);
+                    Settings.Messages = value;
```
jimmyca15, line 140: >Settings.Messages = value [](start = 20, length = 25)

Updating object model directly due to update in configuration sounds like a misstep to me. I would expect that the push model still be integrated with IConfiguration. It just exposes a new way to notify the application that it needs to refresh it's IConfiguration.

examples/DotNetCore/EventHub/EventHubConnection.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        public string EventHubConnectionString { get; set; }
+
+        public string EventHubName { get; set; }
+
+        public string BlobStorageConnectionString { get; set; }
+
+        public string BlobStorageContainerName { get; set; }
+
+        public string EventHubConsumerGroup { get; set; }
```
jimmyca15, line 13: >EventHubConsumerGroup [](start = 22, length = 21)

Will the usage different consumer groups affect the applications event processing? For example, will it only see a subset of total events?

.gitignore
gsgvijay: Removed it. I thought it might be better to include it to prevent accidentally checking in connection strings.

examples/DotNetCore/EventHub/Controllers/HomeController.cs
gsgvijay: Moved it to a different service.

examples/DotNetCore/EventHub/EventContract.cs
gsgvijay: Renamed it.

examples/DotNetCore/EventHub/EventHubConnection.cs
gsgvijay: Each consumer group would get a copy of the event. Consumer groups are the way for user applications to run multiple instances. The number of consumer groups will need to be the same as the number of application instances for each instance to get the notification.

examples/DotNetCore/EventHub/Controllers/HomeController.cs
gsgvijay: Changed to call the refresher instead.

gsgvijay:
> Please remove those generated files

Removed some files from wwwroot. Let me know if we should also remove the js and css files too.

.gitignore
```diff
@@ -331,3 +331,7 @@ ASALocalRun/
 
 # MFractors (Xamarin productivity tool) working folder 
 .mfractor/
+
+# Remove generated files
```
MSGaryWang, line 335: >generated [](start = 9, length = 9)

If these files are generated, why we have them checked in? The client script lib should come from npm dependency at build time.

examples/DotNetCore/EventHub/appsettings.Development.json
```diff
@@ -0,0 +1,9 @@
+{
+  "Logging": {
+    "LogLevel": {
```
MSGaryWang, line 3: >LogLevel [](start = 5, length = 8)

these logging levels are same as appsettings.json

examples/DotNetCore/EventHub/appsettings.json
```diff
@@ -0,0 +1,26 @@
...
+    "EventHubConnection": {
+      "EventHubConnectionString": "",
+      "EventHubName": "",
+      "BlobStorageConnectionString": "",
+      "BlobStorageContainerName": "",
+      "AppConfigConnectionString": ""
+    },
+    "Demo": {
+      "Settings": {
+        "Messages": "Hello there!; General Kenobi! You are a bold one.",
```
MSGaryWang, line 19: >Kenobi [](start = 43, length = 6)

You might only use Microsoft approved name

examples/DotNetCore/EventHub/Views/Shared/_Layout.cshtml
```diff
@@ -0,0 +1,51 @@
...
+
+    <footer class="border-top footer text-muted">
+        <div class="container">
+            &copy; 2020 - SimpleAppConfigEventHub - <a asp-area="" asp-controller="Home" asp-action="Privacy">Privacy</a>
+        </div>
+    </footer>
+    <script src="~/lib/jquery/dist/jquery.min.js"></script>
+    <script src="~/lib/bootstrap/dist/js/bootstrap.bundle.min.js"></script>
+    <script src="~/js/site.js" asp-append-version="true"></script>
+    @RenderSection("Scripts", required: false)
```
MSGaryWang, line 49: why we need these code pieces, we should remove them if it's sample for other purposes

examples/DotNetCore/EventHub/EventHubConnection.cs
```diff
@@ -0,0 +1,17 @@
+namespace SimpleAppConfigEventHub
+{
+    public class EventHubConnection
```
MSGaryWang, line 3: >EventHubConnection [](start = 17, length = 18)

Please have comments for each properties

MSGaryWang:
You would need READM.md to explain everything about this sample project

examples/DotNetCore/EventHub/EventHubService.cs
```diff
@@ -0,0 +1,158 @@
...
+            {
+                var key = e.Data.Key;
+                if (IsWatchedKeyValue(key))
+                {
+                    shouldRefresh = true;
+                    break;
+                }
+            }
+
+            if (shouldRefresh)
```
MSGaryWang, line 138: >shouldRefresh [](start = 16, length = 13)

if user change the key within _kvSelectors and we will refresh every time? we might demo it using single key or delay.

.gitignore
gsgvijay: Removed the changes.

examples/DotNetCore/EventHub/appsettings.json
```diff
@@ -0,0 +1,25 @@
...
+      "Microsoft.Hosting.Lifetime": "Information"
+    }
+  },
+  "AllowedHosts": "*",
+  "EventHubConnection": {
+    "EventHubConnectionString": "Endpoint=sb://appconfig-eh-demo.servicebus.windows.net/;SharedAccessKeyName=policy1;SharedAccessKey=BZOWcz9lSjJTI6rr4RJlQwhXmwenTCLVFR/l4jVtG4Y=;EntityPath=demo-hub-1",
+    "EventHubName": "demo-hub-1",
+    "BlobStorageConnectionString": "DefaultEndpointsProtocol=https;AccountName=ehdemo;AccountKey=WHbB4td6i+lNqqznwpCH/GnQwTURGsNP5vK2JMKu/i6bAkTcfAa2Ae5JZ5TMe0govDOfhcL73i0ymP1QKMjnUA==;EndpointSuffix=core.windows.net",
+    "BlobStorageContainerName": "ehcontainer",
+    "AppConfigConnectionString": "Endpoint=https://vijay-app-config.azconfig.io;Id=0-l3-s0:i9/OWy8WCSF9grwHJTwN;Secret=KJNRaN3KHL6Rzn1im8KCQi/FnGLiMh3ORPwzrN86JAM="
```
MSGaryWang, line 15: Please never put secret into source code

examples/DotNetCore/EventHub/appsettings.json
```diff
@@ -0,0 +1,25 @@
...
+    "LogLevel": {
+      "Default": "Information",
+      "Microsoft": "Warning",
+      "Microsoft.Hosting.Lifetime": "Information"
+    }
+  },
+  "AllowedHosts": "*",
+  "EventHubConnection": {
+    "EventHubConnectionString": "",
+    "EventHubName": "demo-hub-1",
```
zhenlan, line 12: Should this be set by users?

examples/DotNetCore/EventHub/appsettings.json
```diff
@@ -0,0 +1,25 @@
...
+    "BlobStorageConnectionString": "",
+    "BlobStorageContainerName": "ehcontainer",
```
zhenlan, line 14: Should this be set by users?

examples/DotNetCore/EventHub/EventHubService.cs
```diff
@@ -0,0 +1,156 @@
...
+            _processorClient.ProcessEventAsync += ProcessEventHandler;
+            _processorClient.ProcessErrorAsync += ProcessErrorHandler;
+
+            _processorClient.StartProcessing();
+        }
+        
+        private void InitAppConfig()
+        {
+            var builder = new ConfigurationBuilder();
+            builder.AddAzureAppConfiguration(options =>
```
zhenlan, line 61: This is odd. Why are we doing this twice, once here, once in the ctor?

examples/DotNetCore/EventHub/EventHubService.cs
```diff
@@ -0,0 +1,156 @@
...
+        private IEnumerable<string> _kvSelectors = null;
+
+        public EventHubService(IConfiguration configuration)
+        {
+            var builder = new ConfigurationBuilder();
+
+            _eventHubConnection = configuration.GetSection("EventHubConnection").Get<EventHubConnection>();
+            builder.AddAzureAppConfiguration(options =>
+            {
+                options.Connect(configuration["EventHubConnection:AppConfigConnectionString"])
```
zhenlan, line 37: The App Config connection string shouldn't be under EventHubConnection section. Actually, we should never show an example that putting any connection strings in a config file. All secrets should be from the environment.

examples/DotNetCore/EventHub/appsettings.json
```diff
@@ -0,0 +1,25 @@
...
+  "Logging": {
+    "LogLevel": {
+      "Default": "Information",
+      "Microsoft": "Warning",
+      "Microsoft.Hosting.Lifetime": "Information"
+    }
+  },
+  "AllowedHosts": "*",
+  "EventHubConnection": {
+    "EventHubConnectionString": "",
```
zhenlan, line 11: If this is a secret, it shouldn't be here.

examples/DotNetCore/EventHub/appsettings.json
```diff
@@ -0,0 +1,25 @@
...
+    "EventHubName": "demo-hub-1",
+    "BlobStorageConnectionString": "",
```
zhenlan, line 13: This is a secret. It shouldn't be in a config file.

examples/DotNetCore/EventHub/appsettings.json
```diff
@@ -0,0 +1,25 @@
...
+    "BlobStorageContainerName": "ehcontainer",
+    "AppConfigConnectionString": ""
```
zhenlan, line 15: This is a secret. You know how easy one can accidently include an actual secret value in a PR, right?

examples/DotNetCore/EventHub/appsettings.json
```diff
@@ -0,0 +1,25 @@
...
+  },
+  "Demo": {
+    "Settings": {
```
zhenlan, line 18: Why two layers? Can we have "settings" at the root?

examples/DotNetCore/EventHub/appsettings.json
```diff
@@ -0,0 +1,25 @@
+{
+  "Logging": {
+    "LogLevel": {
+      "Default": "Information",
+      "Microsoft": "Warning",
+      "Microsoft.Hosting.Lifetime": "Information"
+    }
+  },
+  "AllowedHosts": "*",
+  "EventHubConnection": {
```
zhenlan, line 10: I have an idea. Can all these settings be load from App Config?

examples/DotNetCore/EventHub/EventHubService.cs
```diff
@@ -0,0 +1,156 @@
...
+
+            foreach (var e in events)
+            {
+                var key = e.Data.Key;
+                if (IsWatchedKeyValue(key))
+                {
+                    shouldRefresh = true;
+                    break;
+                }
+            }
```
zhenlan, line 134: This is over complicated. We should just call `TryRefreshAsync()` if there are incoming events. If users want to scope, they can filter on the event grid side easily. No need to do this runtime checking.

examples/DotNetCore/EventHub/EventHubService.cs
```diff
@@ -0,0 +1,156 @@
...
+
+            if (shouldRefresh)
+            {
+                await _refresher.RefreshAsync();
+                Settings = Configuration.GetSection("Demo:Settings").Get<Settings>();
```
zhenlan, line 139: We shouldn't need to do this. Use `IOptionsSnapshot<Settings>`.
[https://github.com/Azure/AppConfiguration/blob/master/examples/DotNetCore/AzureFunction/FunctionApp/ShowMessage.cs#L17](https://github.com/Azure/AppConfiguration/blob/master/examples/DotNetCore/AzureFunction/FunctionApp/ShowMessage.cs#L17)

examples/DotNetCore/EventHub/EventHubService.cs
```diff
@@ -0,0 +1,156 @@
...
+            }
+        }
+
+        private Task ProcessErrorHandler(ProcessErrorEventArgs eventArgs)
+        {
+            Console.WriteLine($"\n\nERROR: Partition: '{eventArgs.PartitionId}': an unhandled exception was encountered.\n\n");
```
zhenlan, line 145: How bad when this happens? Is it transient (automatic recoverable) or it must need someone's attention?

examples/DotNetCore/EventHub/EventHubService.cs
```diff
@@ -0,0 +1,156 @@
...
+            Console.WriteLine(eventArgs.Exception.Message);
+            return Task.CompletedTask;
+        }
+
+        private bool IsWatchedKeyValue(string key)
```
zhenlan, line 150: No need to do this.

Instead, you should register what key to watch for changes. Please use a sentinel key.

examples/DotNetCore/EventHub/KeyValueEvent.cs
```diff
@@ -0,0 +1,44 @@
...
+        public DateTime EventTime { get; set; }
+
+        [JsonProperty("dataVersion")]
+        public int DataVersion { get; set; }
+
+        [JsonProperty("metadataVersion")]
+        public int MetadataVersion { get; set; }
+    }
+
+    public class Data
```
zhenlan, line 33: Every class should live in a separate file. You may want to give a more meaningful name to this class.

examples/DotNetCore/EventHub/Program.cs
```diff
@@ -0,0 +1,36 @@
...
+            Host.CreateDefaultBuilder(args)
+                .ConfigureWebHostDefaults(webBuilder =>
+                {
+                    webBuilder.ConfigureAppConfiguration(builder =>
+                    {
+                        var settings = builder.Build();
+                        var appConfigConnectionString = settings["EventHubConnection:AppConfigConnectionString"];
+                        if (!string.IsNullOrEmpty(appConfigConnectionString))
+                        {
+                            builder.AddAzureAppConfiguration(options =>
```
zhenlan, line 25: Wow, this is a third place we are doing this. We need to do some cleanup.

examples/DotNetCore/EventHub/SimpleAppConfigEventHub.csproj
```diff
@@ -0,0 +1,25 @@
...
+  <ItemGroup>
+      <PackageReference Include="Azure.Messaging.EventHubs" Version="5.0.1" />
+      <PackageReference Include="Azure.Messaging.EventHubs.Processor" Version="5.0.1" />
+      <PackageReference Include="Microsoft.Extensions.Configuration" Version="3.1.3" />
+      <PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="3.1.3" />
+      <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="3.1.3" />
+      <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
+      <PackageReference Include="Microsoft.Azure.AppConfiguration.AspNetCore" Version="3.0.0" />
+      <PackageReference Include="Azure.Identity" Version="1.1.1" />
+      <PackageReference Include="Microsoft.Extensions.Configuration.AzureAppConfiguration" Version="3.0.0" />
```
zhenlan, line 16: This will be brought in by `Microsoft.Azure.AppConfiguration.AspNetCore` already.

examples/DotNetCore/EventHub/SimpleAppConfigEventHub.csproj
```diff
@@ -0,0 +1,25 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+      <PackageReference Include="Azure.Messaging.EventHubs" Version="5.0.1" />
+      <PackageReference Include="Azure.Messaging.EventHubs.Processor" Version="5.0.1" />
+      <PackageReference Include="Microsoft.Extensions.Configuration" Version="3.1.3" />
+      <PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="3.1.3" />
+      <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="3.1.3" />
+      <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
+      <PackageReference Include="Microsoft.Azure.AppConfiguration.AspNetCore" Version="3.0.0" />
```
zhenlan, line 14: Please use 3.0.1 or 3.0.*

examples/DotNetCore/EventHub/Controllers/HomeController.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using Azure.Storage.Blobs;
+using Azure.Messaging.EventHubs.Consumer;
+using Azure.Messaging.EventHubs.Processor;
+using Newtonsoft.Json;
+using Azure.Messaging.EventHubs;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Configuration.AzureAppConfiguration;
```
zhenlan, line 15: Sort & delete usings

examples/DotNetCore/EventHub/Controllers/HomeController.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace SimpleAppConfigEventHub.Controllers
+{
+    public class HomeController : Controller
+    {
+        private IEventHubService _eventHubService;
+
+        public HomeController(IEventHubService eventHubService)
```
zhenlan, line 23: This has nothing to do with event hub. All it cases is the settings.

zhenlan:
Can we just reuse the existing [WebDemo](https://github.com/Azure/AppConfiguration/tree/master/examples/DotNetCore/WebDemo) app and add Event Hub support? I believe it will simplify the app a lot (and reduce the number of files).

PR #314, opened 2020-04-14
Initial version of ASP.NET (full framework) quickstart

lisaguthrie:


jimmyca15:
Looks good to me. The only suggestion I could have is that we remove anything unneeded from the ASP.Net template. Not sure if there is anything unneeded though.

PR #321, opened 2020-04-27
Spring Release Notes for 1.x.5

mrm9084:


releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -6,6 +6,14 @@
 
 [Source code web ][source_code_web] | [Package (Maven) web][package_web] | [Product documentation][docs]
 
+## 1.1.4/1.2.4 - April 27, 2020
+
+* Add ability to modify client connection to App Configuration and Key Vault. Clients can be provided via ```ConfigurationClientBuilderSetup``` and ```SecretClientBuilderSetup```.
```
abhilasharora, line 11: Suggestion : Added support to allow users to configure the client used to connect to App Configuration and Key Vault

abhilasharora, line 11: >``` [](start = 106, length = 3)

A single backtick could be used for inline code (instead of 3 backticks)

abhilasharora, line 11: >```ConfigurationClientBuilderSetup``` and ```SecretClientBuilderSetup``` [](start = 106, length = 72)

We can be more explicit that users can implement these interfaces to configure the default client.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -6,6 +6,14 @@
...
+  * This allows for setting up additional client information such as proxy information.
```
abhilasharora, line 12: >proxy information [](start = 69, length = 17)

Is there an issue or code sample that we can link here that shows how users can specify the proxy information?

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -6,6 +6,14 @@
...
+* Fixed the bug that caused a partial store to load when both /application/ and /<application_name>/ are used, but one failed to load. This also caused the store to be unable to refresh.
```
abhilasharora, line 13: >/application/ and /<application_name>/ [](start = 62, length = 38)

These two settings can be enclosed in backticks for emphasis.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -6,6 +6,14 @@
...
+* Fixed the bug that caused extra requests to be made when a large number of changes existed in the store history.
+* Fixed the bug where using empty label could not be used and was document as null label, which returned all results.
```
abhilasharora, line 15: Suggestion : Fixed a bug when using an empty string as label that caused key-values with `null` label to be returned instead

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -6,6 +6,14 @@
 
 [Source code web ][source_code_web] | [Package (Maven) web][package_web] | [Product documentation][docs]
 
+## 1.1.4/1.2.4 - April 27, 2020
+
+* Add ability to modify client connection to App Configuration and Key Vault. Clients can be provided via ```ConfigurationClientBuilderSetup``` and ```SecretClientBuilderSetup```.
+  * This allows for setting up additional client information such as proxy information.
+* Fixed the bug that caused a partial store to load when both /application/ and /<application_name>/ are used, but one failed to load. This also caused the store to be unable to refresh.
+* Fixed the bug that caused extra requests to be made when a large number of changes existed in the store history.
```
abhilasharora, line 14: What issue did we fix here? Is there a PR for those changes?

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
zhenlan: Yes, we need to provide more details than this. We can update the document and add a link in the release notes.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
zhenlan: The sentence does not sound right. Please rephrase.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
zhenlan: Not sure I understand. Please rephrase.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
zhenlan: Fixed a bug that may cause the excessive number of requests being made during configuration change detection to the App Configuration.

(Release notes is to help customers to understand whether an issue applies to them. It does not have to reveal all implementation details. Add an related issue here. It can be useful for anyone to learn more about the issue including implementation details.)

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -6,6 +6,14 @@
 
 [Source code web ][source_code_web] | [Package (Maven) web][package_web] | [Product documentation][docs]
 
+## 1.1.4/1.2.4 - April 27, 2020
```
zhenlan, line 9: Please add any related issues.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -6,6 +6,15 @@
...
+
+* Add support to allow users configure the client used to connect to App Configuration and Key Vault. Users can define the clients via `ConfigurationClientBuilderSetup` and `SecretClientBuilderSetup`, shown [here](https://github.com/microsoft/spring-cloud-azure/tree/master/spring-cloud-azure-starters/spring-cloud-starter-azure-appconfiguration-config#modifying-connection-client). [#656](https://github.com/microsoft/spring-cloud-azure/issues/656)
```
abhilasharora, line 11: The first sentence can be updated to past tense to maintain consistency with other points. For example, "Added" instead of "Add".

abhilasharora, line 11: We can use the "to" preposition here.
allow users configure -> allow users to configure

abhilasharora, line 11: This issue seems to be related to the sub-bullet instead of the main bullet.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -6,6 +6,15 @@
...
+  * This allows for setting up additional client information such as proxy information.
+* Fixed the bug that caused a partial store to load when both `/application/` and `/<application_name>/` are used, and only one of them loaded. This also caused the store to be unable to refresh.
```
abhilasharora, line 13: The user might not be importing all settings from the configuration store.
partial store -> partial settings OR partial configuration

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -6,6 +6,15 @@
...
+* Fixed a bug that may cause the excessive number of requests being made during configuration change detection to the App Configuration. [#672](https://github.com/microsoft/spring-cloud-azure/issues/672)
```
abhilasharora, line 14: Current : Fixed a bug that may cause the excessive number of requests being made during configuration change detection to the App Configuration.

Suggestion : Improved the performance of refresh by reducing the number of requests made to App Configuration during configuration change detection.

abhilasharora, line 14: It might be good to respond to this issue before publishing the release notes. We can let the customer know that we have made some improvements in the latest version of the package, and suggest them to upgrade their application to use it.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -6,6 +6,15 @@
...
+* Fixed the bug where using an empty string label could not be done. [#655](https://github.com/microsoft/spring-cloud-azure/issues/655)
```
abhilasharora, line 15: We can be more explicit that we now allow users to specify `\0` to denote an empty string as the value for the label. I don't see this information in the issue that you linked.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -6,6 +6,15 @@
...
+  * Also a documentation issues was fixed that caused confusion between empty and null label, where null label returned all results.
```
abhilasharora, line 16: I'm not sure if we should include this in the release notes.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
mrm9084: I plan on it, but there is an issue with the linked docs we need to fix first.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -6,6 +6,14 @@
 
 [Source code web ][source_code_web] | [Package (Maven) web][package_web] | [Product documentation][docs]
 
+## 1.1.5/1.2.5 - April 27, 2020
+
+* Added support to allow users to configure the client used to connect to App Configuration and Key Vault. Users can now define the clients via `ConfigurationClientBuilderSetup` and `SecretClientBuilderSetup`, shown [here](https://github.com/microsoft/spring-cloud-azure/tree/master/spring-cloud-azure-starters/spring-cloud-starter-azure-appconfiguration-config#modifying-connection-client).
+  * This allows for setting up additional client information such as proxy information. [#656](https://github.com/microsoft/spring-cloud-azure/issues/656)
+* Fixed the bug that caused a partial configuration to load when both `/application/` and `/<application_name>/` are used, and only one of them loaded. This also caused the store to be unable to refresh.
```
abhilasharora, line 13: >caused a partial configuration [](start = 21, length = 30)

caused a partial configuration -> caused partial configuration

abhilasharora, line 13: >* Fixed the bug that caused a partial configuration to load when both `/application/` and `/<application_name>/` are used, and only one of them loaded. This also caused the store to be unable to refresh [](start = 0, length = 202)

"This also caused" in the second sentence is confusing since the first sentence talks about "Fixed the bug". How about the following?

Fixed the bug that caused configuration to load partially and refresh to fail when both `/application/` and `/<application_name>/` were used but only one of them loaded.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -15,7 +23,7 @@
 
 ## 1.1.2/1.2.2 - February 25, 2020
 
-* Credentials for authentication can now be provided in code via ```AppConfigurationCredentialProvider``` and ```KeyVaultCredentialProvider```.
+* Credentials for authentication can now be provided in code via `AppConfigurationCredentialProvider` and `KeyVaultCredentialProvider`.
```
zhenlan, line 26: Do you want to do this for the entire doc?

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -6,6 +6,14 @@
 
 [Source code web ][source_code_web] | [Package (Maven) web][package_web] | [Product documentation][docs]
 
+## 1.1.5/1.2.5 - April 27, 2020
+
+* Added support to allow users to configure the client used to connect to App Configuration and Key Vault. Users can now define the clients via `ConfigurationClientBuilderSetup` and `SecretClientBuilderSetup`, shown [here](https://github.com/microsoft/spring-cloud-azure/tree/master/spring-cloud-azure-starters/spring-cloud-starter-azure-appconfiguration-config#modifying-connection-client).
+  * This allows for setting up additional client information such as proxy information. [#656](https://github.com/microsoft/spring-cloud-azure/issues/656)
```
zhenlan, line 12: * Introduced new interfaces, which allows users to customize client builders for connecting to App Configuration and Key Vault. See this [document](https://github.com/microsoft/spring-cloud-azure/tree/master/spring-cloud-azure-starters/spring-cloud-starter-azure-appconfiguration-config#modifying-connection-client) for more details and examples.  [#656](https://github.com/microsoft/spring-cloud-azure/issues/656)
  * `ConfigurationClientBuilderSetup`
  * `SecretClientBuilderSetup`

zhenlan: BTW, the document is still not updated. We should update the document first.
https://github.com/microsoft/spring-cloud-azure/tree/master/spring-cloud-azure-starters/spring-cloud-starter-azure-appconfiguration-config#modifying-connection-client

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -6,6 +6,14 @@
...
+* Fixed the bug that caused partial configuration to load when both `/application/` and `/<application_name>/` are used, and only one of them loaded. This also caused the store to be unable to refresh.
```
zhenlan, line 13: This sounds like a terrible terrible bug. Something is missing. How partial the config is loaded?

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -6,6 +6,14 @@
...
+* Improved the performance of refresh by reducing the number of requests made to App Configuration during configuration change detection. [#672](https://github.com/microsoft/spring-cloud-azure/issues/672)
```
zhenlan, line 14: By "performance of refresh", are we refreshing faster? Is this the revision pagination issue? If so, I would suggest below:

* Fixed the bug that more than necessary requests may be sent for configuration change detection especially when an App Configuration store has many keys or many revision histories.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -6,6 +6,14 @@
...
+* Fixed the bug where using an empty string label "\0" or a comma without a value before it could not be used. [#655](https://github.com/microsoft/spring-cloud-azure/issues/655)
```
zhenlan, line 15: The "%00" was working. Not sure when it was broken. Using "\0" as null label is a breaking change... at least from a documentation perspective. Suggest below.

* Fixed the bug that `%00` was not working when it's used to indicate keys with no labels explicitly. Changed to use `\0` to indicate keys with no labels. For example, a property setting below means loading keys with no labels and then overwritten by keys with label `dev`. 
    ```properties
    spring.cloud.azure.appconfiguration.stores[0].label = \0,dev
    ```

mrm9084: I looked into it and it looks like there was a slight change the label check when it switched to using the Azure sdk.

zhenlan: Yes, I suspect this was broken during that time too.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -6,6 +6,15 @@
...
+  * `ConfigurationClientBuilderSetup`
+  * `SecretClientBuilderSetup`
+* Fixed the bug that caused configuration to load partially and refresh to fail when both `/application/` and `/<application_name>/` were used but only one of them loaded.
+* Fixed the bug that more than necessary requests may be made for configuration change detection especially when an App Configuration store has many keys or many change revisions. [#672](https://github.com/microsoft/spring-cloud-azure/issues/672)
+* Fixed the bug that `%00` was not working when it's used to indicate keys with no labels explicitly. Changed to use `\0` to indicate keys with no labels to be consistent with the service. For example, a property setting below means loading keys with no labels and then overwritten by keys with label `dev`. [#655](https://github.com/microsoft/spring-cloud-azure/issues/655)
```
zhenlan, line 16: Where is the example?

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -6,6 +6,15 @@
 
 [Source code web ][source_code_web] | [Package (Maven) web][package_web] | [Product documentation][docs]
 
+## 1.1.5/1.2.5 - April 27, 2020
+
+* Introduced new interfaces, which allows users to customize client builders for connecting to App Configuration and Key Vault. See this [document](https://github.com/microsoft/spring-cloud-azure/tree/master/spring-cloud-azure-starters/spring-cloud-starter-azure-appconfiguration-config#modifying-connection-client) for more details and examples. [#656](https://github.com/microsoft/spring-cloud-azure/issues/656)
+  * `ConfigurationClientBuilderSetup`
+  * `SecretClientBuilderSetup`
+* Fixed the bug that caused configuration to load partially and refresh to fail when both `/application/` and `/<application_name>/` were used but only one of them loaded.
```
zhenlan, line 14: How partially? Is there a bug I can find more details?

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -6,6 +6,15 @@
 
 [Source code web ][source_code_web] | [Package (Maven) web][package_web] | [Product documentation][docs]
 
+## 1.1.5/1.2.5 - April 27, 2020
+
+* Introduced new interfaces, which allows users to customize client builders for connecting to App Configuration and Key Vault. See this [document](https://github.com/microsoft/spring-cloud-azure/tree/master/spring-cloud-azure-starters/spring-cloud-starter-azure-appconfiguration-config#modifying-connection-client) for more details and examples. [#656](https://github.com/microsoft/spring-cloud-azure/issues/656)
```
zhenlan, line 11: ... which allow~s~ users...

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
mrm9084: Partially means that if if you had 1 config store that used both /application/ and /application_dev/ if one failed to lead the other could still load. I don't think this bug was ever reported as we saw it as a result of the ms docs issue with key vault.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -6,6 +6,19 @@
...
+  * `ConfigurationClientBuilderSetup`
+  * `SecretClientBuilderSetup`
+* Fixed the bug that caused configuration to load partially and refresh to fail when both `/application/` and `/<application_name>/` were used but only one of them loaded.
+* Fixed the bug that more than necessary requests may be made for configuration change detection especially when an App Configuration store has many keys or many change revisions. [#672](https://github.com/microsoft/spring-cloud-azure/issues/672)
+* Fixed the bug that `%00` was not working when it's used to indicate keys with no labels explicitly. Changed to use `\0` to indicate keys with no labels to be consistent with the service. For example, a property setting below means loading keys with no labels and then overwritten by keys with label `dev`. [#655](https://github.com/microsoft/spring-cloud-azure/issues/655)
+
+```java
+spring.cloud.azure.appconfiguration.stores[0].label = \0,dev
+```
```
zhenlan, line 20: Add empty spaces in front so the indentation aligns with the bullet point above.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
zhenlan: I see. I guess the fix is not to load everything (instead of partial) but to drop whatever that has been loaded. We should make it clear. Does this only happen when fail-fast is false?

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -6,6 +6,19 @@
 
 [Source code web ][source_code_web] | [Package (Maven) web][package_web] | [Product documentation][docs]
 
+## 1.1.5/1.2.5 - April 27, 2020
+
+* Introduced new interfaces, which allow users to customize client builders for connecting to App Configuration and Key Vault. See this [document](https://github.com/microsoft/spring-cloud-azure/tree/master/spring-cloud-azure-starters/spring-cloud-starter-azure-appconfiguration-config#modifying-connection-client) for more details and examples. [#656](https://github.com/microsoft/spring-cloud-azure/issues/656)
```
zhenlan, line 11: The link needs to be updated.

mrm9084: Do you mean the reply to the issue?

zhenlan: No, the link to the document.

mrm9084: The link has been updated.

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
mrm9084: Yes, it only happened with fail-fast set to false. And the change is that the whole store loads or nothing is loaded.

zhenlan: We should make it clear for things like this in the future. Here is what I would suggest this time:

When fail-fast is disabled, an application may only load configuration under one context but not the other from an App Configuration store. The bug is fixed so either all configuration is loaded successfully or nothing is loaded to ensure the consistency of configuration in the application.

PR #323, opened 2020-05-04
Add release note for vstsTask 3.0.13

Yiming-Jia:
* The latest release removed the deployment of `pipeline-task-lib` to bypass misencoded symbol issue.

### DIFF
--- a/releaseNotes/AzureDevOpsPipelineExtension.md
+++ b/releaseNotes/AzureDevOpsPipelineExtension.md
@@ -5,7 +5,7 @@ The Azure App Configuration extension in Azure DevOps pipeline has been publishe
 ![sample](pictures/AzureDevOpsExtensionVersionSample.PNG)
 
 ### v3.0.13 - May, 04 2020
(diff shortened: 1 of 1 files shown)


releaseNotes/AzureDevOpsPipelineExtension.md
```diff
@@ -4,6 +4,9 @@ The Azure App Configuration extension in Azure DevOps pipeline has been publishe
 
 ![sample](pictures/AzureDevOpsExtensionVersionSample.PNG)
 
+### v3.0.13 - May, 04 2020
+* Fixed the issue that some reserved symbols misencoded set to environment variables. [#313](https://github.com/Azure/AppConfiguration/issues/313)
```
zhenlan, line 8: We don't set environment variables. How about this?

Fixed the issue that some symbols, for example, percent sign (%), could not be set to Pipeline Task variables correctly.

Yiming-Jia: Yes, should try to avoid `task variable` or `environment variable`.

PR #329, opened 2020-05-13
Added event hub integration

gsgvijay:


examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Pages/Index.cshtml.cs
```diff
@@ -1,13 +1,12 @@
 ﻿namespace WebDemoWithEventHub.Pages
 {
     using Microsoft.AspNetCore.Mvc.RazorPages;
-    using Microsoft.Extensions.Options;
 
     public class IndexModel : PageModel
     {
-        public IndexModel(IOptionsSnapshot<Settings> options)
+        public IndexModel(ISettingsProvider settingsProvider)
```
zhenlan, line 7: We only change how the application detects for config changes (from poll model to push model). This shouldn't change how the application consumes configuration. I don't expect this file changes at all.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/WebDemoWithEventHub.csproj
```diff
@@ -3,8 +3,13 @@
...
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.AppConfiguration.AspNetCore" Version="3.0.0" />
+    <PackageReference Include="Microsoft.Azure.AppConfiguration.AspNetCore" Version="3.0.*" />
     <PackageReference Include="Azure.Identity" Version="1.1.1" />
+    <PackageReference Include="Azure.Messaging.EventHubs" Version="5.0.1" />
+    <PackageReference Include="Azure.Messaging.EventHubs.Processor" Version="5.0.1" />
+    <PackageReference Include="Microsoft.Extensions.Configuration" Version="3.1.3" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="3.1.3" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="3.1.3" />
```
zhenlan, line 12: As we discussed, we shouldn't need to include the 3 packages above. They will be brought in automatically by dependencies.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
```diff
@@ -0,0 +1,107 @@
...
+        private EventHubConnection eventHubConnection;
+
+        private EventProcessorClient processorClient;
+
+        private IConfigurationRefresher refresher;
+
+        public EventHubService(IConfiguration config, IOptions<Settings> options)
+        {
+            var builder = new ConfigurationBuilder();
+            builder.AddAzureAppConfiguration(options =>
```
zhenlan, line 29: As we discussed, you shouldn't do this again.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
```diff
@@ -0,0 +1,107 @@
...
+
+            processorClient.ProcessEventAsync += ProcessEventHandler;
+            processorClient.ProcessErrorAsync += ProcessErrorHandler;
+            processorClient.StartProcessing();
+        }
+
+        private async Task ProcessEventHandler(ProcessEventArgs eventArgs)
+        {
+            await refresher.TryRefreshAsync();
+            settings = configuration.GetSection("WebDemo:Settings").Get<Settings>();
```
zhenlan, line 97: We already do this in startup.cs. You don't need to do this again here.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
```diff
@@ -0,0 +1,107 @@
...
+                                                       eventHubName: eventHubConnection.EventHubName);
+
+            processorClient.ProcessEventAsync += ProcessEventHandler;
+            processorClient.ProcessErrorAsync += ProcessErrorHandler;
+            processorClient.StartProcessing();
+        }
+
+        private async Task ProcessEventHandler(ProcessEventArgs eventArgs)
+        {
+            await refresher.TryRefreshAsync();
```
zhenlan, line 96: I think we can refresh asynchronously here. Remove `await`.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/ISettingsProvider.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace WebDemoWithEventHub
+{
+    public interface ISettingsProvider
```
zhenlan, line 3: We shouldn't need this interface.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
```diff
@@ -0,0 +1,107 @@
...
+
+        private IConfiguration configuration;
+
+        private EventHubConnection eventHubConnection;
+
+        private EventProcessorClient processorClient;
+
+        private IConfigurationRefresher refresher;
+
+        public EventHubService(IConfiguration config, IOptions<Settings> options)
```
zhenlan, line 26: We need to figure out how to get the instance of `IConfigurationRefresher` through DI, not `IConfiguration`.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Pages/Index.cshtml.cs
gsgvijay: Undid this change.

---
In reply to: [426103473](https://github.com/Azure/AppConfiguration/pull/329#discussion_r426103473) [](ancestors = 426103473)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/WebDemoWithEventHub.csproj
gsgvijay: Removed these.

---
In reply to: [426103640](https://github.com/Azure/AppConfiguration/pull/329#discussion_r426103640) [](ancestors = 426103640)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: Removed these. We're adding AzureAppConfiguration only once now.

---
In reply to: [426103973](https://github.com/Azure/AppConfiguration/pull/329#discussion_r426103973) [](ancestors = 426103973)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: Removed await.

---
In reply to: [426104145](https://github.com/Azure/AppConfiguration/pull/329#discussion_r426104145) [](ancestors = 426104145)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: Added another class to inject the IConfigurationRefresher.

---
In reply to: [426104383](https://github.com/Azure/AppConfiguration/pull/329#discussion_r426104383) [](ancestors = 426104383)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/ISettingsProvider.cs
gsgvijay: Removed it.

---
In reply to: [426104206](https://github.com/Azure/AppConfiguration/pull/329#discussion_r426104206) [](ancestors = 426104206)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: Removed it.

---
In reply to: [426104072](https://github.com/Azure/AppConfiguration/pull/329#discussion_r426104072) [](ancestors = 426104072)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/IConfigRefresher.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace WebDemoWithEventHub
+{
+    public interface IConfigRefresher
```
zhenlan, line 3: We already have `IConfigurationRefresher`. Why do we need to introduce a new interface?

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Program.cs
```diff
@@ -39,12 +49,17 @@ public static void Main(string[] args)
...
+                                        .ConfigureKeyVault(kv =>
+                                        {
+                                            kv.SetCredential(new DefaultAzureCredential());
+                                        })
                                        .Select(keyFilter: "WebDemo:*")
                                        .ConfigureRefresh((refreshOptions) =>
                                        {
-                                           // Indicates that all configuration should be refreshed when the given key has changed.
-                                           refreshOptions.Register(key: "WebDemo:Sentinel", refreshAll: true);
+                                           refreshOptions.Register(key: "WebDemo:Settings", refreshAll: true);
```
zhenlan, line 59: Why is this change?

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Program.cs
```diff
@@ -39,12 +49,17 @@ public static void Main(string[] args)
                             builder.AddAzureAppConfiguration(options =>
                             {
                                 options.Connect(new Uri(appConfigurationEndpoint), new DefaultAzureCredential())
+                                        .ConfigureKeyVault(kv =>
+                                        {
+                                            kv.SetCredential(new DefaultAzureCredential());
+                                        })
```
zhenlan, line 55: Please move this after the `.ConfigureRefresh` call.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Program.cs
```diff
@@ -2,15 +2,25 @@
...
     {
+        private static IConfigurationRefresher refresher = null;
+
         public static void Main(string[] args)
         {
-            CreateHostBuilder(args).Build().Run();
+            IHostBuilder builder = CreateHostBuilder(args);
+            builder.ConfigureServices((services) =>
+            {
+                services.AddSingleton<IConfigRefresher>(new ConfigRefresher() { Refresher = refresher });
```
zhenlan, line 20: Is this this the right place? Should this be in the startup.cs?

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
```diff
@@ -0,0 +1,86 @@
...
+                                                       consumerGroup: consumerGroup,
+                                                       connectionString: _eventHubConnection.EventHubConnectionString,
+                                                       eventHubName: _eventHubConnection.EventHubName);
+
+            _processorClient.ProcessEventAsync += ProcessEventHandler;
+            _processorClient.ProcessErrorAsync += ProcessErrorHandler;
+            _processorClient.StartProcessing();
+        }
+
+        private Task ProcessEventHandler(ProcessEventArgs eventArgs)
```
zhenlan, line 73: Can this function be async?

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
```diff
@@ -0,0 +1,86 @@
...
+        private EventProcessorClient _processorClient;
+
+        private IConfigRefresher _configRefresher;
+
+        public EventHubService(IConfiguration config, IConfigRefresher configRefresher)
+        {
+            _configRefresher = configRefresher;
+
+            _eventHubConnection = config.GetSection("WebDemo:Connection").Get<EventHubConnection>();
+            _settings = config.GetSection("WebDemo:Settings").Get<Settings>();
```
zhenlan, line 29: We shouldn't need this.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        private EventProcessorClient _processorClient;
+
+        private IConfigRefresher _configRefresher;
+
+        public EventHubService(IConfiguration config, IConfigRefresher configRefresher)
+        {
+            _configRefresher = configRefresher;
+
+            _eventHubConnection = config.GetSection("WebDemo:Connection").Get<EventHubConnection>();
```
zhenlan, line 28: I would name it "WebDemo:EventHubConnection" to match what you have in code.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
```diff
@@ -0,0 +1,86 @@
...
+            _settings = config.GetSection("WebDemo:Settings").Get<Settings>();
+
+            InitEventHubProcessor();
+        }
+
+        private void InitEventHubProcessor()
```
zhenlan, line 34: The `_eventHubConnection` is only used in this function. We can make it a parameter passed this function.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
```diff
@@ -0,0 +1,86 @@
...
+
+namespace WebDemoWithEventHub
+{
+    public class EventHubService
+    {
+        private Settings _settings;
+
+        private EventHubConnection _eventHubConnection;
+
+        private EventProcessorClient _processorClient;
```
zhenlan, line 20: Why does this need to be a member variable?

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
```diff
@@ -0,0 +1,86 @@
...
+using Microsoft.Extensions.Configuration.AzureAppConfiguration;
+using Microsoft.Extensions.Options;
+using System;
+using System.Threading.Tasks;
+
+namespace WebDemoWithEventHub
+{
+    public class EventHubService
+    {
+        private Settings _settings;
```
zhenlan, line 16: We don't need this.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/IConfigRefresher.cs
gsgvijay: We would need an interface for DI, right? Without this, ConfigRefresher.cs class would need to implement the IConfigurationExplorer

---
In reply to: [427590458](https://github.com/Azure/AppConfiguration/pull/329#discussion_r427590458) [](ancestors = 427590458)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Program.cs
gsgvijay: Undid this.

---
In reply to: [427590805](https://github.com/Azure/AppConfiguration/pull/329#discussion_r427590805) [](ancestors = 427590805)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Program.cs
gsgvijay: Moved and renamed.

---
In reply to: [427591529](https://github.com/Azure/AppConfiguration/pull/329#discussion_r427591529) [](ancestors = 427591529)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: It can be async, but it doesn't await on anything, right? This method could have a "void" type return statement but the handler method expects a Task.

---
In reply to: [427594593](https://github.com/Azure/AppConfiguration/pull/329#discussion_r427594593) [](ancestors = 427594593)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: Removed.

---
In reply to: [427595227](https://github.com/Azure/AppConfiguration/pull/329#discussion_r427595227) [](ancestors = 427595227)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: Renamed.

---
In reply to: [427596123](https://github.com/Azure/AppConfiguration/pull/329#discussion_r427596123) [](ancestors = 427596123)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: Changed it to an arg.

---
In reply to: [427597726](https://github.com/Azure/AppConfiguration/pull/329#discussion_r427597726) [](ancestors = 427597726)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: Removed it.

---
In reply to: [427598190](https://github.com/Azure/AppConfiguration/pull/329#discussion_r427598190) [](ancestors = 427598190)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: Changed to local variable.

---
In reply to: [427598084](https://github.com/Azure/AppConfiguration/pull/329#discussion_r427598084) [](ancestors = 427598084)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Program.cs
gsgvijay: It needs the "refresher" instance which is only available here.

---
In reply to: [427593552](https://github.com/Azure/AppConfiguration/pull/329#discussion_r427593552) [](ancestors = 427593552)

gsgvijay: I've updated it to get the "Refresher" instance from the IConfiguration.Providers list.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
```diff
@@ -71,11 +63,7 @@ private void InitEventHubProcessor(EventHubConnection eventHubConnection)
 
         private Task ProcessEventHandler(ProcessEventArgs eventArgs)
         {
-            foreach (var refresher in _refreshers)
-            {
-                refresher.TryRefreshAsync();
-            }
-
+            _configRoot?.Reload();
```
abhilasharora, line 66: Just to add some context for reviewers, this `Reload` method is an existing method that will reconstruct the configuration from all configuration providers. @drago-draganov and I had a discussion and we feel that this would be a good starting point for a Reload method to include in our initial examples for consuming event grid events. We may plan to implement a similar method specific to our configuration provider in the future based on customer feedback.

zhenlan: I could check but may have the answer faster :) ... Is `Reload` only available on `IConfigurationRoot` not `IConfiguration`?

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Program.cs
```diff
@@ -42,8 +42,11 @@ public static void Main(string[] args)
                                        .Select(keyFilter: "WebDemo:*")
                                        .ConfigureRefresh((refreshOptions) =>
                                        {
-                                           // Indicates that all configuration should be refreshed when the given key has changed.
-                                           refreshOptions.Register(key: "WebDemo:Sentinel", refreshAll: true);
+                                           refreshOptions.Register(key: "WebDemo:*", refreshAll: true);
```
zhenlan, line 45: This won't work. We don't support watching on *. Why is the change?

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
zhenlan: I don't see the code change.

---
In reply to: [428450341](https://github.com/Azure/AppConfiguration/pull/329#discussion_r428450341) [](ancestors = 428450341,427596123)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
zhenlan: If events of 10 config changes all come in a very short period of time, are we going to reload the whole configuration (ours and others) 10 times? We throw the caching and the sentinel key (for configuration consistency) all out of the window?

abhilasharora: Yes, `Reload` method is available only in the derived class `IConfigurationRoot` but not in the base class `IConfiguration`.

---
In reply to: [432184784](https://github.com/Azure/AppConfiguration/pull/329#discussion_r432184784) [](ancestors = 432184784)

abhilasharora: Here's one way to look at this approach.

1. Pull and push refresh are de-coupled from each other. The `ConfigureRefresh` method is used to enable pull refresh and it allows the user to limit the settings to monitor as a part of refresh. Push refresh is based on events that pass the filter criteria for the event grid subscription. Receipt of such an event means that the user's intent is to update either one specific settings that corresponds to the event or all selected settings.

2. Users can configure their event grid subscription to raise events only for their sentinel key. A change in the sentinel key would result in Reload for their configuration in a consistent manner. So users can update any number of settings and then change the sentinel key to reload their configuration. What is the maximum number of *sentinel* changes we expect the users to make in the 99th percentile? Is it 10 or lower?

3. Whether we choose to de-couple pull and push refresh or not, users need to configure their event grid subscription with the appropriate filters. If we get 1000 events from event grid for different key-value changes, but the user has configured refresh only for a single key that was never updated, do we really want to allow users to invoke pull-based Refresh in this case?

---
In reply to: [432188646](https://github.com/Azure/AppConfiguration/pull/329#discussion_r432188646) [](ancestors = 432188646)

zhenlan: There a number of hypotheses and "users need to"s here. I am not looking for a workaround. I am looking for a solution that is simple and intuitive.

This may not be the right place for this discussion. We should compare pros and cons if we go on a totally different route for push-updates.

gsgvijay: I agree with Abhilash that we could allow users to only listen for the sentinel key and refresh from that, but EventGrid/EventHub also helps us deal with issues of too many config changes in a short period of time. The event handler isn't invoked for each config change, but rather, the "ProcessEventArgs" argument contains a list of config changes since the last time event hub polled for changes. The exact polling interval isn't document, but based on my observation, it seems that there's usually a few seconds gap between each invocation of the event handler.

---
In reply to: [432612934](https://github.com/Azure/AppConfiguration/pull/329#discussion_r432612934) [](ancestors = 432612934,432188646)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Program.cs
gsgvijay: I see. But the IConfigurationRoot's Refresh method still did refresh the entire configuration. If we don't have a regex matching in the provider, then I'm not sure if this line is doing anything.

---
In reply to: [432185410](https://github.com/Azure/AppConfiguration/pull/329#discussion_r432185410) [](ancestors = 432185410)

abhilasharora: `IConfigurationRoot.Reload` doesn't use the settings specified for refresh using `ConfigureRefresh`. In case we decide to continue with this approach, then users don't need to register specific settings for refresh in their code. For other alternatives that don't use `Reload`, these settings might be needed.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Program.cs
```diff
@@ -39,11 +39,14 @@ public static void Main(string[] args)
                             builder.AddAzureAppConfiguration(options =>
                             {
                                 options.Connect(new Uri(appConfigurationEndpoint), new DefaultAzureCredential())
-                                       .Select(keyFilter: "WebDemo:*")
```
zhenlan, line 42: Why delete this?

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Program.cs
```diff
@@ -39,11 +39,14 @@ public static void Main(string[] args)
...
+
                                        .ConfigureRefresh((refreshOptions) =>
                                        {
-                                           // Indicates that all configuration should be refreshed when the given key has changed.
```
zhenlan, line 45: Why delete the comment?

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Startup.cs
```diff
@@ -51,5 +54,10 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
                 endpoints.MapRazorPages();
             });
         }
+
+        private void InitServices()
+        {
+            new EventHubService(Configuration);
```
zhenlan, line 60: Is this in the mercy of GC to be deleted?

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
```diff
@@ -0,0 +1,77 @@
...
+{
+    public class EventHubService
+    {
+        IConfigurationRoot _configRoot;
+
+        public EventHubService(IConfiguration config)
+        {
+            EventHubConnection eventHubConnection = config.GetSection("WebDemo:Connection").Get<EventHubConnection>();
+            InitEventHubProcessor(eventHubConnection);
+            if (config is IConfigurationRoot)
```
zhenlan, line 19: What if this is false?

We should log error here. Use ILogger.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
```diff
@@ -0,0 +1,77 @@
...
+        private Task ProcessEventHandler(ProcessEventArgs eventArgs)
+        {
+            _configRoot?.Reload();
+            return Task.CompletedTask;
+        }
+
+        private Task ProcessErrorHandler(ProcessErrorEventArgs eventArgs)
+        {
+            Console.WriteLine($"\n\nERROR: Partition: '{eventArgs.PartitionId}': an unhandled exception was encountered.\n\n");
+            Console.WriteLine(eventArgs.Exception.Message);
```
zhenlan, line 73: Instead of console, I think we should use ILogger.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
```diff
@@ -0,0 +1,77 @@
...
+                                                       eventHubName: eventHubConnection.EventHubName);
+
+            processorClient.ProcessEventAsync += ProcessEventHandler;
+            processorClient.ProcessErrorAsync += ProcessErrorHandler;
+            processorClient.StartProcessing();
+        }
+
+        private Task ProcessEventHandler(ProcessEventArgs eventArgs)
+        {
+            _configRoot?.Reload();
```
zhenlan, line 66: All right, this will wait for the new API.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: Switched to ILogger.

---
In reply to: [433597291](https://github.com/Azure/AppConfiguration/pull/329#discussion_r433597291) [](ancestors = 433597291)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: Redid the change.

---
In reply to: [432186230](https://github.com/Azure/AppConfiguration/pull/329#discussion_r432186230) [](ancestors = 432186230,428450341,427596123)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: Added a logger.

---
In reply to: [433597177](https://github.com/Azure/AppConfiguration/pull/329#discussion_r433597177) [](ancestors = 433597177)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Program.cs
gsgvijay: Reset the change.

---
In reply to: [433595900](https://github.com/Azure/AppConfiguration/pull/329#discussion_r433595900) [](ancestors = 433595900)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Program.cs
gsgvijay: Reset change.

---
In reply to: [433595829](https://github.com/Azure/AppConfiguration/pull/329#discussion_r433595829) [](ancestors = 433595829)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: Used the Cache Invalidation approach we discussed.

---
In reply to: [432641835](https://github.com/Azure/AppConfiguration/pull/329#discussion_r432641835) [](ancestors = 432641835)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Startup.cs
gsgvijay: Injected it instead, and warmed it up in the Configure() method.

---
In reply to: [433596607](https://github.com/Azure/AppConfiguration/pull/329#discussion_r433596607) [](ancestors = 433596607)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            EventHubConnection eventHubConnection = config.GetSection("WebDemo:EventHubConnection").Get<EventHubConnection>();
+            InitEventHubProcessor(eventHubConnection);
+            if (config is IConfigurationRoot configRoot)
+            {
+                foreach (var provider in configRoot.Providers)
+                {
+                    if (provider is Microsoft.Extensions.Configuration.AzureAppConfiguration.IConfigurationRefresher)
+                    {
+                        _refresher = (IConfigurationRefresher)provider;
```
zhenlan, line 31: The refresher instance can now be obtained via DI

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
```diff
@@ -0,0 +1,99 @@
...
+            processorClient.StartProcessing();
+        }
+
+        private Task ProcessEventHandler(ProcessEventArgs eventArgs)
+        {
+            _logger.LogInformation("EventHub update received. Triggering cache invalidation.");
+
+            //
+            // Invalidate cache to force refresh.
+            _refresher?.ResetCache();
```
zhenlan, line 86: Replace with the new SetDirty API

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Startup.cs
```diff
@@ -22,6 +22,7 @@ public void ConfigureServices(IServiceCollection services)
 
             // We add a Settings model to the service container, which takes its values from the applications configuration.
             services.Configure<Settings>(Configuration.GetSection("WebDemo:Settings"));
+            services.AddSingleton<EventHubService>();
```
zhenlan, line 25: Please add a comment to explain this call

zhenlan:
@gsgvijay now that the [v4.0.0-preview](https://www.nuget.org/packages/Microsoft.Extensions.Configuration.AzureAppConfiguration/4.0.0-preview) config provider is released, please update the code accordingly.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Startup.cs
gsgvijay: Added comment.

---
In reply to: [462050387](https://github.com/Azure/AppConfiguration/pull/329#discussion_r462050387) [](ancestors = 462050387)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: Updated it.

---
In reply to: [462049933](https://github.com/Azure/AppConfiguration/pull/329#discussion_r462049933) [](ancestors = 462049933)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: Updated this to get the refresher instance through DI.

---
In reply to: [462049704](https://github.com/Azure/AppConfiguration/pull/329#discussion_r462049704) [](ancestors = 462049704)

gsgvijay:
Updated the reference version.

---
In reply to: [665444188](https://github.com/Azure/AppConfiguration/pull/329#issuecomment-665444188) [](ancestors = 665444188)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
```diff
@@ -0,0 +1,88 @@
...
+
+namespace WebDemoWithEventHub
+{
+    public class EventHubService
+    {
+        private IConfigurationRefresher _refresher;
+
+        private ILogger<EventHubService> _logger;
+
+        public EventHubService(IConfiguration config, IConfigurationRefresherProvider providers, ILogger<EventHubService> logger)
```
zhenlan, line 20: This is an example. Let's use good naming for variables. Suggest `providers` -> `refresherProvider`.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
```diff
@@ -0,0 +1,88 @@
...
+        {
+            _logger = logger;
+
+            EventHubConnection eventHubConnection = config.GetSection("WebDemo:EventHubConnection").Get<EventHubConnection>();
+            InitEventHubProcessor(eventHubConnection);
+
+            _refresher = providers.Refreshers.FirstOrDefault(r => r is IConfigurationRefresher);
```
zhenlan, line 27: Why `r => r is IConfigurationRefresher`?

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
```diff
@@ -0,0 +1,88 @@
...
+            processorClient.StartProcessing();
+        }
+
+        private Task ProcessEventHandler(ProcessEventArgs eventArgs)
+        {
+            _logger.LogInformation("EventHub update received. Triggering cache invalidation.");
+
+            //
+            // Invalidate cache to force refresh.
+            _refresher?.SetDirty(maxDelay: TimeSpan.FromSeconds(1));
```
zhenlan, line 75: Don't override the default value in the example.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
```diff
@@ -0,0 +1,88 @@
...
+            processorClient.ProcessErrorAsync += ProcessErrorHandler;
+            processorClient.StartProcessing();
+        }
+
+        private Task ProcessEventHandler(ProcessEventArgs eventArgs)
+        {
+            _logger.LogInformation("EventHub update received. Triggering cache invalidation.");
+
+            //
+            // Invalidate cache to force refresh.
```
zhenlan, line 74: The comment is technically incorrect. Can you copy it from the `SetDirty` API?

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Startup.cs
```diff
@@ -22,6 +23,11 @@ public void ConfigureServices(IServiceCollection services)
 
             // We add a Settings model to the service container, which takes its values from the applications configuration.
             services.Configure<Settings>(Configuration.GetSection("WebDemo:Settings"));
+
+            services.AddAzureAppConfiguration();
```
zhenlan, line 27: Please add a comment

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Startup.cs
```diff
@@ -22,6 +23,11 @@ public void ConfigureServices(IServiceCollection services)
 
             // We add a Settings model to the service container, which takes its values from the applications configuration.
             services.Configure<Settings>(Configuration.GetSection("WebDemo:Settings"));
```
zhenlan, line 25: Should `EventHubConnection` be configured in the same way so it can be obtained via DI by `EventHubServcie`?

gsgvijay: Changed it to a DI.

---
In reply to: [464788619](https://github.com/Azure/AppConfiguration/pull/329#discussion_r464788619) [](ancestors = 464788619)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Startup.cs
gsgvijay: Added comment from the package.

---
In reply to: [464787723](https://github.com/Azure/AppConfiguration/pull/329#discussion_r464787723) [](ancestors = 464787723)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: Copied from the API.

---
In reply to: [464787625](https://github.com/Azure/AppConfiguration/pull/329#discussion_r464787625) [](ancestors = 464787625)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: refresherProvider.Refreshers is an IEnumerable. I'm trying to assign the first IConfigurationRefresher instance to _configuartionRerfresher.

---
In reply to: [464787077](https://github.com/Azure/AppConfiguration/pull/329#discussion_r464787077) [](ancestors = 464787077)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: Updated.

---
In reply to: [464786843](https://github.com/Azure/AppConfiguration/pull/329#discussion_r464786843) [](ancestors = 464786843)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: Removed this.

---
In reply to: [464787236](https://github.com/Azure/AppConfiguration/pull/329#discussion_r464787236) [](ancestors = 464787236)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Startup.cs
```diff
@@ -1,6 +1,7 @@
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Configuration.AzureAppConfiguration;
```
zhenlan, line 4: Why is this needed?

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Startup.cs
```diff
@@ -20,8 +21,15 @@ public void ConfigureServices(IServiceCollection services)
         {
             services.AddRazorPages();
 
-            // We add a Settings model to the service container, which takes its values from the applications configuration.
+            // We add a Settings and EventHubConnection models to the service container, which takes its values from the applications configuration.
             services.Configure<Settings>(Configuration.GetSection("WebDemo:Settings"));
+            services.Configure<EventHubConnection>(Configuration.GetSection("WebDemo:EventHubConnection"));
+
+            // Adds Azure App Configuration services to the specified Microsoft.Extensions.DependencyInjection.IServiceCollection.
```
zhenlan, line 28: Let's follow the existing style. How about?

"Add Azure App Configuration required services to the DI".

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
zhenlan: Understood. My question is why the lambda is needed.

---
In reply to: [466011776](https://github.com/Azure/AppConfiguration/pull/329#discussion_r466011776) [](ancestors = 466011776,464787077)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
zhenlan: Can we do `IOptionsSnapshot<EventHubConnection>`?

---
In reply to: [466011871](https://github.com/Azure/AppConfiguration/pull/329#discussion_r466011871) [](ancestors = 466011871,464786843)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Startup.cs
gsgvijay: Updated the comment.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/Startup.cs
gsgvijay: Removed unused import.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: Changed to IOptions.

IOptionsSnapshot only works with Scoped services, and doesn't work with Singleton. And we need EventHubService to be singleton to activate it in Startup.cs.

---
In reply to: [466103302](https://github.com/Azure/AppConfiguration/pull/329#discussion_r466103302) [](ancestors = 466103302,466011871,464786843)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: I thought it would be simpler. Would a forloop be clearer to understand?

---
In reply to: [466102871](https://github.com/Azure/AppConfiguration/pull/329#discussion_r466102871) [](ancestors = 466102871,466011776,464787077)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
```diff
@@ -0,0 +1,90 @@
...
+            }
+
+            string consumerGroup = string.IsNullOrEmpty(eventHubConnection.EventHubConsumerGroup)
+                ? EventHubConsumerClient.DefaultConsumerGroupName
+                : eventHubConnection.EventHubConsumerGroup;
+
+            var storageClient = new BlobContainerClient(connectionString: eventHubConnection.StorageConnectionString,
+                                                        blobContainerName: eventHubConnection.StorageContainerName);
+
+            var processorClient = new EventProcessorClient(checkpointStore: storageClient,
```
steverash, line 61: I think processorClient needs to be stored in a field.   This local variable gets GC'd and none of the events seem to fire.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
zhenlan: Won't just `FirstOrDefault()` be sufficient?

---
In reply to: [466813876](https://github.com/Azure/AppConfiguration/pull/329#discussion_r466813876) [](ancestors = 466813876,466102871,466011776,464787077)

gsgvijay: Changed to FirstOrDefault()

---
In reply to: [467235445](https://github.com/Azure/AppConfiguration/pull/329#discussion_r467235445) [](ancestors = 467235445,466813876,466102871,466011776,464787077)

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
gsgvijay: Added a field.

---
In reply to: [467140570](https://github.com/Azure/AppConfiguration/pull/329#discussion_r467140570) [](ancestors = 467140570)

PR #360, opened 2020-07-02
Update release notes for .NET Provider v3.0.2

abhilasharora:
This pull request updates the release notes for version `3.0.2` of the package `Microsoft.Extensions.Configuration.AzureAppConfiguration`.

### DIFF
--- a/releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
+++ b/releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
@@ -1,6 +1,9 @@
 ## Microsoft.Extensions.Configuration.AzureAppConfiguration
 ### [Package (NuGet)](https://www.nuget.org/packages/Microsoft.Extensions.Configuration.AzureAppConfiguration)
 
+### 3.0.2 - July 01, 2020
+* Fixed the issue that caused a configuration change to be skipped as a part of refresh when a change is detected in a key with `refreshAll` set to `true` and all attempts to retrieve the updated settings from App Configuration fail. [#178](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/178)
+
 ### 3.0.1 - April 02, 2020
 * Improved refresh for feature flags to reduce the number of calls made to App Configuration if no change is detected. [#138](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/138)
 * Fixed the issue that caused `TryRefreshAsync` to throw an `Azure.Identity.AuthenticationFailedException` when the `TokenCredential` used to fetch a key vault reference failed to authenticate. [#149](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/149)
--- a/releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
+++ b/releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
@@ -2,7 +2,7 @@
 ### [Package (NuGet)](https://www.nuget.org/packages/Microsoft.Extensions.Configuration.AzureAppConfiguration)
 
 ### 3.0.2 - July 01, 2020
-* Fixed the issue that caused a configuration change to be skipped as a part of refresh when a change is detected in a key with `refreshAll` set to `true` and all attempts to retrieve the updated settings from App Configuration fail. [#178](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/178)
+* Fixed an issue that may cause configuration refresh to be ignored when the key registered to refresh all configuration has changed. [#178](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/178)
 
 ### 3.0.1 - April 02, 2020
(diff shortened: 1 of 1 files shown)


releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
```diff
@@ -1,6 +1,9 @@
 ## Microsoft.Extensions.Configuration.AzureAppConfiguration
 ### [Package (NuGet)](https://www.nuget.org/packages/Microsoft.Extensions.Configuration.AzureAppConfiguration)
 
+### 3.0.2 - July 01, 2020
+* Fixed the issue that caused a configuration change to be skipped as a part of refresh when a change is detected in a key with `refreshAll` set to `true` and all attempts to retrieve the updated settings from App Configuration fail. [#178](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/178)
+
```
zhenlan, line 6: Are there other fixes. I remember you mentioned there were other fixes made previously but not released.

abhilasharora: The only additional change included in this version is to update the user agent header policy to be invoked once per call instead of once per retry. I don't think users care about it, so I have excluded it from the release notes.

jimmyca15: It's still handy to put down even if it's just for our reference. One thing I've liked about the release notes is being able to pinpoint a version a user on a previous version is using and seeing all of the behavior differences between their version and the latest.

zhenlan: One way of doing it is to create milestones for each release, which includes all changes in a release, even a fix not mentioned in release notes as it's not of too much customer interests (but we can always include a link of the milestone in the release notes). This is also beneficial for our planning. For example, we can create a v4.0.0 milestone. We will feel clearer what have done and what is still remaining before we can release it.

jimmyca15: Milestones aren't a bad idea, but I think they are an added value on top of a complete release notes document. They're kind of a GitHub concept, where as release notes really could be moved anywhere.

As far as mentioning the minor fix, I would include it in the release notes but I'm fine if we don't.

releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
```diff
@@ -1,6 +1,9 @@
 ## Microsoft.Extensions.Configuration.AzureAppConfiguration
 ### [Package (NuGet)](https://www.nuget.org/packages/Microsoft.Extensions.Configuration.AzureAppConfiguration)
 
+### 3.0.2 - July 01, 2020
+* Fixed the issue that caused a configuration change to be skipped as a part of refresh when a change is detected in a key with `refreshAll` set to `true` and all attempts to retrieve the updated settings from App Configuration fail. [#178](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/178)
```
zhenlan, line 5: Feels like too much implementation detail. How about this?

"Fixed an issue that configuration may not be refreshed even though the key registered to refresh all configuration has changed."

releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
abhilasharora: We currently have a copy of the release notes at two places - one in this file and the other on the release page [here](https://github.com/Azure/AppConfiguration-DotnetProvider/releases). Currently, both of these are mostly the same, but we can add additional information to one of these if that helps.

releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
abhilasharora: Sure, we can reduce the details. I think it might still be helpful to include that it only affects scenarios when `refreshAll:true` is set for a key. This will allow users who do not use it to know that they weren't impacted. How about this?

Fixed an issue that may cause configuration to not refresh on change in a key with `refreshAll: true`

zhenlan: I was trying to be descriptive. "the key registered to refresh all configuration" is basically `refreshAll: true`. However, the latter is out of context. It's not clear what API we are talking about here. We cannot assume customers always know the context as us. You do have a linked issue that has sufficient implementation details, so I think a descriptive note works better in this case.

I see you prefer the style of "... issue that caused ...". I don't know which one is better :) Jimmy can help. Here is another version:

"Fixed an issue that may cause configuration refresh to be ignored even though the key registered to refresh all configuration has changed."

releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
zhenlan: Let's not complicate this. Keep them the same so it's simpler to maintain. I am fine without mentioning user agent string fix in release notes.

PR #191, opened 2019-11-16
Added Version info and links

mrm9084:

PR #378, opened 2020-07-24
Add release notes for configuration provider v4.0.0-preview

abhilasharora:
This pull request adds the release notes for the .NET Configuration Provider v4.0.0-preview.
https://www.nuget.org/packages/Microsoft.Extensions.Configuration.AzureAppConfiguration/4.0.0-preview
https://www.nuget.org/packages/Microsoft.Azure.AppConfiguration.AspNetCore/4.0.0-preview

### DIFF
--- a/releaseNotes/MicrosoftAzureAppConfigurationAspNetCore.md
+++ b/releaseNotes/MicrosoftAzureAppConfigurationAspNetCore.md
@@ -3,9 +3,12 @@
 
 ### 4.0.0-preview - July 23, 2020
 * Added multi-targeting support for .NET Core 3.1 besides .NET Standard 2.0. [#173](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/173)
-* Extracted out the code to retrieve required dependencies for configuration refresh from `IApplicationBuilder.UseAzureAppConfiguration()` to `IServiceCollection.AddAzureAppConfiguration()` in the package `Microsoft.Extensions.Configuration.AzureAppConfiguration`, so it could be re-used for non ASP.NET Core applications.
-    * Users must call `IServiceCollection.AddAzureAppConfiguration()` to register the required services for configuration refresh before they can call `IApplicationBuilder.UseAzureAppConfiguration()`.
+
+* Updated code to leverage the new feature of dependency injection support for obtaining `IConfigurationRefresher` instances introduced in the `4.0.0-preview` version of the `Microsoft.Extensions.Configuration.AzureAppConfiguration` package.
+    * Users must call `IServiceCollection.AddAzureAppConfiguration()` in `ConfigureServices(...)` to register the required services for configuration refresh before they can call `IApplicationBuilder.UseAzureAppConfiguration()`. This makes it easier to retrieve instances of `IConfigurationRefresher` through dependency injection in a controller or a middleware, and have better control of when and how configuration is refreshed.
+
     * An exception is thrown when the required services for configuration refresh could not be retrieved from the `IServiceCollection` instance. [#166](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/166)
+
 * Updated `Microsoft.Extensions.Configuration.AzureAppConfiguration` reference to `4.0.0-preview`. See the [release notes](./MicrosoftExtensionsConfigurationAzureAppConfiguration.md) for more information on the changes.
 
 ### 3.0.2 - July 01, 2020
--- a/releaseNotes/MicrosoftAzureAppConfigurationAspNetCore.md
+++ b/releaseNotes/MicrosoftAzureAppConfigurationAspNetCore.md
@@ -2,7 +2,10 @@
 ### [Package (NuGet)](https://www.nuget.org/packages/Microsoft.Azure.AppConfiguration.AspNetCore)
 
 ### 4.0.0-preview - July 23, 2020
-* Added multi-targeting support for .NET Core 3.1. [#173](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/173)
+* Added multi-targeting support for .NET Core 3.1 besides .NET Standard 2.0. [#173](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/173)
+* Extracted out the code to retrieve required dependencies for configuration refresh from `IApplicationBuilder.UseAzureAppConfiguration()` to `IServiceCollection.AddAzureAppConfiguration()` in the package `Microsoft.Extensions.Configuration.AzureAppConfiguration`, so it could be re-used for non ASP.NET Core applications.
+    * Users must call `IServiceCollection.AddAzureAppConfiguration()` to register the required services for configuration refresh before they can call `IApplicationBuilder.UseAzureAppConfiguration()`.
+    * An exception is thrown when the required services for configuration refresh could not be retrieved from the `IServiceCollection` instance. [#166](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/166)
 * Updated `Microsoft.Extensions.Configuration.AzureAppConfiguration` reference to `4.0.0-preview`. See the [release notes](./MicrosoftExtensionsConfigurationAzureAppConfiguration.md) for more information on the changes.
 
 ### 3.0.2 - July 01, 2020
--- a/releaseNotes/MicrosoftAzureAppConfigurationAspNetCore.md
+++ b/releaseNotes/MicrosoftAzureAppConfigurationAspNetCore.md
@@ -4,12 +4,12 @@
 ### 4.0.0-preview - July 23, 2020
 * Added multi-targeting support for .NET Core 3.1 besides .NET Standard 2.0. [#173](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/173)
 
-* Updated code to leverage the new feature of dependency injection support for obtaining `IConfigurationRefresher` instances introduced in the `4.0.0-preview` version of the `Microsoft.Extensions.Configuration.AzureAppConfiguration` package.
+* **Breaking Change :** To leverage the new feature of dependency injection support for obtaining `IConfigurationRefresher` instances introduced in the `4.0.0-preview` version of the `Microsoft.Extensions.Configuration.AzureAppConfiguration` package, the following changes are made.
     * Users must call `IServiceCollection.AddAzureAppConfiguration()` in `ConfigureServices(...)` to register the required services for configuration refresh before they can call `IApplicationBuilder.UseAzureAppConfiguration()`. This makes it easier to retrieve instances of `IConfigurationRefresher` through dependency injection in a controller or a middleware, and have better control of when and how configuration is refreshed.
 
     * An exception is thrown when the required services for configuration refresh could not be retrieved from the `IServiceCollection` instance. [#166](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/166)
 
-* Updated `Microsoft.Extensions.Configuration.AzureAppConfiguration` reference to `4.0.0-preview`. See the [release notes](./MicrosoftExtensionsConfigurationAzureAppConfiguration.md) for more information on the changes.
+* **Breaking Change :** Updated `Microsoft.Extensions.Configuration.AzureAppConfiguration` reference to `4.0.0-preview`. See the [release notes](./MicrosoftExtensionsConfigurationAzureAppConfiguration.md) for more information on the changes.
 
 ### 3.0.2 - July 01, 2020
 * Updated `Microsoft.Extensions.Configuration.AzureAppConfiguration` reference to `3.0.2`. See the [release notes](./MicrosoftExtensionsConfigurationAzureAppConfiguration.md) for more information on the changes.
--- a/releaseNotes/MicrosoftAzureAppConfigurationAspNetCore.md
+++ b/releaseNotes/MicrosoftAzureAppConfigurationAspNetCore.md
@@ -1,6 +1,10 @@
 ## Microsoft.Azure.AppConfiguration.AspNetCore
 ### [Package (NuGet)](https://www.nuget.org/packages/Microsoft.Azure.AppConfiguration.AspNetCore)
 
+### 4.0.0-preview - July 23, 2020
+* Added multi-targeting support for .NET Core 3.1. [#173](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/173)
+* Updated `Microsoft.Extensions.Configuration.AzureAppConfiguration` reference to `4.0.0-preview`. See the [release notes](./MicrosoftExtensionsConfigurationAzureAppConfiguration.md) for more information on the changes.
+
 ### 3.0.2 - July 01, 2020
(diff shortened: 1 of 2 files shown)


releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
```diff
@@ -1,6 +1,14 @@
 ## Microsoft.Extensions.Configuration.AzureAppConfiguration
 ### [Package (NuGet)](https://www.nuget.org/packages/Microsoft.Extensions.Configuration.AzureAppConfiguration)
 
+### 4.0.0-preview - July 23, 2020
+* Added `IConfigurationRefresher.SetDirty` method  to set the cached value for key-values registered for refresh as dirty. For users that use push notifications from event grid to receive configuration change updates, this allows the cached values to be revalidated on the next call to `RefreshAsync` or `TryRefreshAsync`.
+* Added support for deserializing configuration settings with JSON content type. [#191](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/191)
```
zhenlan, line 6: How about this?

Added JSON content-type (eg., MIME type `application/json`) support for key-values in App Configuration, so primitive types, arrays, and JSON objects can be loaded to `IConfiguration` properly.

releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
```diff
@@ -1,6 +1,14 @@
...
+* Added an interface `IConfigurationRefresherProvider` to provide access to instances of `IConfigurationRefresher` through dependency injection. This can be configured by adding a call to the following extension method on the service collection. [#167](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/167)
+   ```` csharp
+   public static IServiceCollection AddAzureAppConfiguration(this IServiceCollection services)
+   ````
+
```
zhenlan, line 11: Looks like https://github.com/Azure/AppConfiguration-DotnetProvider/issues/166 is missed from the release notes. It's a breaking change. We should call it out and explain why.

releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
```diff
@@ -1,6 +1,14 @@
 ## Microsoft.Extensions.Configuration.AzureAppConfiguration
 ### [Package (NuGet)](https://www.nuget.org/packages/Microsoft.Extensions.Configuration.AzureAppConfiguration)
 
+### 4.0.0-preview - July 23, 2020
+* Added `IConfigurationRefresher.SetDirty` method  to set the cached value for key-values registered for refresh as dirty. For users that use push notifications from event grid to receive configuration change updates, this allows the cached values to be revalidated on the next call to `RefreshAsync` or `TryRefreshAsync`.
```
zhenlan, line 5: I will reverse the order. The introduction of the API is not the reason. It's just the result of our support for the push model. How about this?

Added enhanced support for applications that leverage Event Grid integration in App Configuration for configuration refresh with push notifications of configuration changes. The new API below is introduced in `IConfigurationRefresher` interface, which can be called when an application responds to events from Event Grid. It will signal the application to reassess whether configuration should be updated upon the next call to `RefreshAsync()` or `TryRefreshAsync()`.
```cs
void SetDirty(TimeSpan? maxDelay = null)
```

zhenlan: BTW, we should add a link to issue https://github.com/Azure/AppConfiguration-DotnetProvider/issues/133.

releaseNotes/MicrosoftAzureAppConfigurationAspNetCore.md
```diff
@@ -1,6 +1,10 @@
 ## Microsoft.Azure.AppConfiguration.AspNetCore
 ### [Package (NuGet)](https://www.nuget.org/packages/Microsoft.Azure.AppConfiguration.AspNetCore)
 
+### 4.0.0-preview - July 23, 2020
+* Added multi-targeting support for .NET Core 3.1. [#173](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/173)
```
zhenlan, line 5: How about this?

Added multi-targeting support for .NET Core 3.1 *besides .NET Standard 2.0*.

releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
```diff
@@ -1,6 +1,14 @@
...
+* Added support for deserializing configuration settings with JSON content type. [#191](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/191)
+* Added an interface `IConfigurationRefresherProvider` to provide access to instances of `IConfigurationRefresher` through dependency injection. This can be configured by adding a call to the following extension method on the service collection. [#167](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/167)
+   ```` csharp
+   public static IServiceCollection AddAzureAppConfiguration(this IServiceCollection services)
+   ````
```
zhenlan, line 10: We should give a full picture before diving into specific APIs. How about this?

Added support for applications to obtain `IConfigurationRefresher` instances from the dependency injection (DI), so applications have better control of when to call `RefreshAsync()`/`TryRefreshAsync()` or whether to `await` the call to complete.

The following two APIs are introduced. To take advantage of this feature, applications need to call `IServiceCollection.AddAzureAppConfiguration()` first to add required services to the DI and then `IConfigurationRefresher` instances are accessible via `IConfigurationRefresherProvider` obtained through DI.
```cs
public static IServiceCollection AddAzureAppConfiguration(this IServiceCollection services)
```
```cs
public interface IConfigurationRefresherProvider
```

releaseNotes/MicrosoftAzureAppConfigurationAspNetCore.md
```diff
@@ -1,6 +1,10 @@
...
+* Updated `Microsoft.Extensions.Configuration.AzureAppConfiguration` reference to `4.0.0-preview`. See the [release notes](./MicrosoftExtensionsConfigurationAzureAppConfiguration.md) for more information on the changes.
+
```
zhenlan, line 7: We have a breaking change that users must call `IServiceCollection.AddAzureAppConfiguration()` before they can call `UseAzureAppConfiguration`. We need to call this out.

releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
avanigupta: Sounds good.

releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
abhilasharora: I will include this in the release notes for the AspNetCore package. We can skip it here since the breaking change in this issue wasn't part of this package before the current release.

zhenlan: That's right.

releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
```diff
@@ -1,6 +1,26 @@
 ## Microsoft.Extensions.Configuration.AzureAppConfiguration
 ### [Package (NuGet)](https://www.nuget.org/packages/Microsoft.Extensions.Configuration.AzureAppConfiguration)
 
+### 4.0.0-preview - July 23, 2020
+* Added enhanced support for applications that leverage Event Grid integration in App Configuration for configuration refresh. The following new API is introduced in `IConfigurationRefresher` interface, which can be called when an application responds to push notifications from Event Grid. This signals the application to reassess whether configuration should be updated on the next call to `RefreshAsync()` or `TryRefreshAsync()`. [#133](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/133)
+   ````csharp
+   void SetDirty(TimeSpan? maxDelay = null)
+   ````
+* Added JSON content-type (e.g. MIME type `application/json`) support for key-values in App Configuration. This allows primitive types, arrays, and JSON objects to be loaded properly to IConfiguration. [#191](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/191)
```
zhenlan, line 9: Backtick IConfiguration

releaseNotes/MicrosoftAzureAppConfigurationAspNetCore.md
```diff
@@ -1,6 +1,13 @@
 ## Microsoft.Azure.AppConfiguration.AspNetCore
 ### [Package (NuGet)](https://www.nuget.org/packages/Microsoft.Azure.AppConfiguration.AspNetCore)
 
+### 4.0.0-preview - July 23, 2020
+* Added multi-targeting support for .NET Core 3.1 besides .NET Standard 2.0. [#173](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/173)
+* Extracted out the code to retrieve required dependencies for configuration refresh from `IApplicationBuilder.UseAzureAppConfiguration()` to `IServiceCollection.AddAzureAppConfiguration()` in the package `Microsoft.Extensions.Configuration.AzureAppConfiguration`, so it could be re-used for non ASP.NET Core applications.
```
zhenlan, line 6: What you said is not wrong, but that's the implementation detail. How about this?

**Breaking Change:** To leverage the new feature of dependency injection support for obtaining `IConfigurationRefresher` instances introduced in the 4.0.0-preview version of the `Microsoft.Extensions.Configuration.AzureAppConfiguration` package, the following changes are made.

releaseNotes/MicrosoftAzureAppConfigurationAspNetCore.md
```diff
@@ -1,6 +1,13 @@
...
+    * Users must call `IServiceCollection.AddAzureAppConfiguration()` to register the required services for configuration refresh before they can call `IApplicationBuilder.UseAzureAppConfiguration()`.
```
zhenlan, line 7: Added a bit more details:

Users must call `IServiceCollection.AddAzureAppConfiguration()` in `ConfigureServices(...)` to register the required services for configuration refresh before they can call `IApplicationBuilder.UseAzureAppConfiguration()`. With this, users can obtain instances of `IConfigurationRefresher` easily from DI in a controller or a middleware and have better control of when and how configuration should be refreshed.

releaseNotes/MicrosoftAzureAppConfigurationAspNetCore.md
```diff
@@ -1,6 +1,16 @@
 ## Microsoft.Azure.AppConfiguration.AspNetCore
 ### [Package (NuGet)](https://www.nuget.org/packages/Microsoft.Azure.AppConfiguration.AspNetCore)
 
+### 4.0.0-preview - July 23, 2020
+* Added multi-targeting support for .NET Core 3.1 besides .NET Standard 2.0. [#173](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/173)
+
+* Updated code to leverage the new feature of dependency injection support for obtaining `IConfigurationRefresher` instances introduced in the `4.0.0-preview` version of the `Microsoft.Extensions.Configuration.AzureAppConfiguration` package.
```
zhenlan, line 7: Any reasons you don't want to call out these are breaking changes explicitly?

Can we rephrase the "Updated code" to something else? You know everything is changing code. It sounds odd.

abhilasharora: We have multiple other breaking changes in this release and in the past. It would be confusing for readers to only see this one changed marked explicitly.

How about the following?
"Updated `IServiceCollection.AddAzureAppConfiguration()` to leverage ..."

zhenlan: What other breaking changes? If we were not clear enough in the past, we should be now. It shouldn't be the reason keeping us from doing it right.



> How about the following?
> "Updated `IServiceCollection.AddAzureAppConfiguration()` to leverage ..."

No, you have details in later bullet points. This will be redundant.

abhilasharora: > What other breaking changes?

I believe the change to add JSON content-type support and the change to add `AppConfigurationEndpoint` to `IConfigurationRefresher` are also breaking changes. Do we want to explicitly specify it for all of these? I skipped it since I already get that information from the major version bump, so it seemed redundant.

> No, you have details in later bullet points. This will be redundant.

An alternative could be "Updated middleware to leverage ...". We already refer to the middleware in the release notes for version `2.0.0-preview-009200001-7`

zhenlan: JSON content-type support falls back to string when something is wrong. It is, however, still possible a potential breaking change. If Avani can come up with scenarios, we should call them out.

The interface changes are technically breaking changes, they do not require users to update their code though. I'm fine you call it out.

Major version change does not mean everything will be broken. That's why it's important to call out what are. Does it make sense?

People will not go back and read what we mean by "middleware". We should have more explanation of what it is if we use the term.

abhilasharora: For users that already use a valid JSON content type for their settings, this is a breaking change. We can call that out to avoid any unexpected behavior changes.

I will check the list of issues with the tag `breaking change` in the milestone, and mark them explicitly in the release notes. This will ensure that we are consistent in our approach of deciding whether a change is breaking or not.

avanigupta: Yes, JSON support is a breaking change if customers are already storing their values as valid JSON with valid JSON content type. For example, if users have this in their store:
```
key=Foo
value=[1,2,3]
contentType=application/json
```
Currently, they might be accessing this setting in their application like `config["Foo"]`. But after the new changes,  `config["Foo"]` will be null. They will now have to access `config["Foo:0"], config["Foo:1"] and config["Foo:2"]`

abhilasharora: What options do these users have to upgrade to the new package and let their applications continue to run as expected, other than manually updating all such settings in App Configuration?

avanigupta: The only option is to update settings in AppConfiguration or the application code itself.

abhilasharora: How about something like this? We can add more details to the referenced issue, if needed.

* **Breaking Change :** Added JSON content-type (e.g. MIME type `application/json`) support for key-values in App Configuration. This allows primitive types, arrays, and JSON objects to be loaded properly to `IConfiguration`. Existing applications that use key-values with a valid JSON content-type may need to be updated. [#191](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/191)

zhenlan: Looks good.

PR #414, opened 2020-10-01
Add R example of hmac signing

actuarialvoodoo:
The hmac.md file has examples in many languages but not one in R. This pull requests provides one. Thank you for having some many language examples, I found python and bash examples helpful in creating one that work with R.

I have included fully qualified function calls (including package) for reference of potential users. 
It would be possible to remove the lubridate dependency and create the `utc_now` value using base R, if that's desirable, but it would be several additional lines of code.


zhenlan:
Closing as the REST API documents have been moved to https://docs.microsoft.com/en-us/azure/azure-app-configuration/rest-api-authentication-hmac.

PR #415, opened 2020-10-05
Update hmac.md for signing request using react native

fawaz-ahmed:
This PR will introduce a way for react native apps to utilize azure app configuration using REST with hmac signing.
Updated the hmac.md file to use crypto and base64 to get the relevant http headers and fetch app config or perform any other operation.

### DIFF
--- a/docs/REST/authentication/hmac.md
+++ b/docs/REST/authentication/hmac.md
@@ -217,12 +217,12 @@ function signRequest(host,
 }
 ```
 
-### React native - Javacript
+### React native
 *Prerequisites*: [RNSimpleCrypto](https://github.com/ghbutton/react-native-simple-crypto), [base64](https://github.com/eranbo/react-native-base64)
 
 ```js
 const signRequest = async (
-  host,             // sample '<xconfig>.azconfig.io'
+  host,             // sample '<config>.azconfig.io'
   method = 'GET',   // GET, PUT, POST, DELETE
   url = '/kv?api-version=1.0',  // path+query
   body = undefined, // request body (undefined of none)
--- a/docs/REST/authentication/hmac.md
+++ b/docs/REST/authentication/hmac.md
@@ -217,6 +217,52 @@ function signRequest(host,
 }
 ```
 
+### React native - Javacript
+*Prerequisites*: [RNSimpleCrypto](https://github.com/ghbutton/react-native-simple-crypto), [base64](https://github.com/eranbo/react-native-base64)
+
+```js
+const signRequest = async (
+  host,             // sample '<xconfig>.azconfig.io'
+  method = 'GET',   // GET, PUT, POST, DELETE
+  url = '/kv?api-version=1.0',  // path+query
+  body = undefined, // request body (undefined of none)
+  credential,       // access key id
+  secret            // access key value (base64 encoded)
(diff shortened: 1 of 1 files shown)


fawazahmed-yara (approved):

docs/REST/authentication/hmac.md
```diff
@@ -217,6 +217,52 @@ function signRequest(host,
 }
 ```
 
+### React native
+*Prerequisites*: [RNSimpleCrypto](https://github.com/ghbutton/react-native-simple-crypto), [base64](https://github.com/eranbo/react-native-base64)
```
abhilasharora, line 221: Thanks for your contribution! Could we use `crypto-js` library instead of `RNSimpleCrypto`? The former is a well-known library with 4000+ public usages as compared to the latter that does not seem to have any public usage.

Link : https://www.npmjs.com/package/crypto-js

docs/REST/authentication/hmac.md
```diff
@@ -217,6 +217,52 @@ function signRequest(host,
 }
 ```
 
+### React native
```
abhilasharora, line 220: We have a one code sample per programming language in the wiki page. Have you tried using the following JavaScript code sample in your React Native application?

https://github.com/Azure/AppConfiguration/blob/main/docs/REST/authentication/hmac.md#javascript

docs/REST/authentication/hmac.md
fawaz-ahmed: @abhilasharora `crypto-js` library is not compatible with `react-native` which is why I had to use another library in place.
The port of `crypto-js` for react native is [`react-native-crypto-js`](https://github.com/imchintan/react-native-crypto-js), but it doesn't supports hmac256 (see it's readme)
Other libs available are:
- [react-native-crypto](https://github.com/tradle/react-native-crypto) the reason for not using this package is that it has a lot of dependency installations before you can use. Refer to installation in readme. Another valid reason is the list of issues filed on github
- [react-native-fast-crypto](https://github.com/EdgeApp/react-native-fast-crypto) doesn't supports hmac - [see issues](https://github.com/EdgeApp/react-native-fast-crypto/issues/14)
- [react-native-crypto](https://github.com/trackforce/react-native-crypto) has issues with pod install - [see issues](https://github.com/trackforce/react-native-crypto/issues/14)
- [react-native-crypto](https://github.com/ghbutton/react-native-simple-crypto) this is a fork of [react-native-crypto](https://github.com/trackforce/react-native-crypto) but it's updated and works great without issues.

As for `base64`, since `crypto-js` was incompatible I had to use another [library](https://github.com/eranbo/react-native-base64) available in place which is pretty standard.

docs/REST/authentication/hmac.md
fawaz-ahmed: Yes I tried. Since the javascript library for `crypto-js` had to be replaced so I had to update as per the available methods in the new package.
The javascript sample you provide shall work for `nodejs` and `browser`, but for `react-native` it needs another variant.

zhenlan:
Closing as the REST API documents have been moved to https://docs.microsoft.com/en-us/azure/azure-app-configuration/rest-api-authentication-hmac.

PR #446, opened 2021-01-05
Update the Azure Functions example to use AppConfig as a config source

zhenlan:
- By taking advantage of the Azure Functions `IFunctionsConfigurationBuilder`, the AppConfig provider can now be added as an additional config source. The Azure Functions existing providers will be preserved. 
  image
- Removed the use of strong type `Settings` and `IOptionsSnapshot` for simplicity. Technically, it's not an AppConfig specific usage pattern. Now that the AppConfig is just one of the providers, users can always do it if it's desired.
- Used the latest API in the config provider v4.1.0 to obtain the refresher instance from the DI
- Updated all dependencies to the latest version


jimmyca15 (approved):
Awesome.

PR #476, opened 2021-03-10
Updated Release Notes for Spring

mrm9084:


AlexandraKemperMS (approved):

releaseNotes/SpringCloudAzureAppConfigurationConfig.md
```diff
@@ -6,6 +6,10 @@
 
 [Source code web ][source_code_web] | [Package (Maven) web][package_web] | [Product documentation][docs]
 
+## 1.2.8 - March 9, 2020
+
+* Updated to newer versions of dependencies. Spring Boot 2.3.5.RELEASE and Spring Cloud 2.2.5.RELEASE.
```
zhenlan, line 11: Looks like the version is simply called 'Spring Boot 2.3.5' instead of 'Spring Boot 2.3.5_.RELEASE_'
https://spring.io/blog/2020/10/29/spring-boot-2-3-5-available-now

While we updated the dependency, I suppose it does not mean customers have to use the new spring versions, correct? If so, we probably should say we added support for the new version.

PR #480, opened 2021-03-12
updated release notes with the latest release

Anne56njeri:


releaseNotes/AzureDevOpsPipelineExtension.md
```diff
@@ -4,6 +4,9 @@ The Azure App Configuration extension in Azure DevOps pipeline has been publishe
 
 ![sample](pictures/AzureDevOpsExtensionVersionSample.PNG)
 
+### v3.4.16 - March, 12 2021
+* Fixed a bug that was causing an unauthorized error when Managed Identities based authentication is used.
```
jimmyca15, line 8: I don't think we ever supported managed identity based authentication. I would think it should be "Added the capability to use managed identity based authentication."

Anne56njeri: I see, makes sense

jimmyca15 (approved):

PR #492, opened 2021-03-23
Add a queue triggered sample function

zhenlan:
Add a queue-triggered sample function with queue name defined in Azure App Configuration.

### DIFF
--- a/examples/DotNetCore/AzureFunction/FunctionApp/ReadQueuedMessage.cs
+++ b/examples/DotNetCore/AzureFunction/FunctionApp/ReadQueuedMessage.cs
@@ -5,7 +5,9 @@ namespace FunctionApp
 {
     public static class ReadQueuedMessage
     {
-        // Queue triggered function with queue name defined in Azure App Configuration
+        // Queue triggered function with queue name defined in Azure App Configuration.
+        // The app setting binding expression is used. See details at
+        // https://docs.microsoft.com/azure/azure-functions/functions-bindings-expressions-patterns
         [FunctionName("ReadQueuedMessage")]
         public static void Run(
             [QueueTrigger(queueName: "%TestApp:Storage:QueueName%")] string myQueueItem, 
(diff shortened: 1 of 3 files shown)


zhenlan:
App Config is already added to the Function App's configuration in the startup. So there is nothing really special to App Config in the sample, but to show what's possible.

examples/DotNetCore/AzureFunction/FunctionApp/ReadQueuedMessage.cs
```diff
@@ -0,0 +1,17 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace FunctionApp
+{
+    public static class ReadQueuedMessage
+    {
+        // Queue triggered function with queue name defined in Azure App Configuration
+        [FunctionName("ReadQueuedMessage")]
+        public static void Run(
+            [QueueTrigger(queueName: "%TestApp:Storage:QueueName%")] string myQueueItem,
```
jimmyca15, line 11: I can't find any documentation on the `%settingName%` syntax. Were you able to? I'd like to have a link to it because at first it's a bit hard to understand what's happening here.

zhenlan: Here is the doc. It's the binding expression of app settings. Do you think I should include the link in the comment?
https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-expressions-patterns#binding-expressions---app-settings

jimmyca15: I think adding the link in a comment would be great.

jimmyca15 (approved):

examples/DotNetCore/AzureFunction/FunctionApp/ReadQueuedMessage.cs
zhenlan: Updated. Good suggestion. Thanks!

PR #504, opened 2021-04-14
Update .NET config provider release notes for package 4.3.0-preview

avanigupta:


releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
```diff
@@ -1,6 +1,22 @@
...
 
+### 4.3.0-preview - April 14,2021
+* Added two new APIs for filtering and trimming feature flag by a prefix. [#234](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/234)
+
+   * Selectively load feature flags based on key and label filters:
+   ```csharp 
+   FeatureFlagOptions Select(string featureFlagFilter, string labelFilter)
+   ```
+
+   * Trim a prefix from the keys of all feature flags retrieved from Azure App Configuration:
```
jimmyca15, line 12: The prefix is trimmed from the [`id` of the feature flag](https://github.com/Azure/AppConfiguration-DotnetProvider/blob/6e880fa7061233b3362f76d5d37923788cc2a18b/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/FeatureManagement/FeatureManagementKeyValueAdapter.cs#L39)

avanigupta: updated

jimmyca15: I see now that the comment in the source code uses "key" as well. We probably want to update that.

jimmyca15 (approved):

PR #575, opened 2021-11-18
JS feature flag web demo

diberry:


diberry:
@AlexandraKemperMS @Muksvso

Muksvso (approved):
Looks good.

zhenlan:
Thanks for the contribution, @diberry.

While this may work as a demo, it has a few issues
- Every request to this app will cause a call to App Config. I'm concerned if someone takes this code and uses it on their server. This is not something we would recommend. Caching is critical on the server.
- The example ignored the filters completely. When a feature flag has filters, its state is determined by the filters not just the enabled property.

We should also have a brief introduction of the example in https://github.com/Azure/AppConfiguration/blob/main/examples/README.md.

examples/JavaScript/WebDemoFeatureFlag/azure-app-configuration.js
```diff
@@ -0,0 +1,31 @@
+const { AppConfigurationClient } = require("@azure/app-configuration");
+
+if (!process.env.production) {
+  require("dotenv").config();
+  console.log("debug environment");
+}
+
+const connectionString = process.env.APPCONFIG_CONNECTION_STRING;
+console.log(connectionString)
```
zhenlan, line 9: Please don't log the connection string. Not a good practice to log secrets.

diberry: @zhenlan Removed, please check the PR now

diberry:
@zhenlan Can you please merge this code if you approve?

zhenlan:
Discussed offline. Due to [issues mentioned earlier](https://github.com/Azure/AppConfiguration/pull/575#issuecomment-973494353), we would not recommend customers to use code like this in any production service. Closing.

PR #588, opened 2021-12-17
Update the example to use the middleware for Azure Functions isolated

zhenlan:


avanigupta (approved):

examples/DotNetCore/AzureFunction/FunctionAppIsolatedMode/FunctionAppIsolatedMode.csproj
```diff
@@ -5,10 +5,10 @@
     <OutputType>Exe</OutputType>
   </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.AppConfiguration.Functions.Worker" Version="5.0.0-preview" />
```
avanigupta, line 8: Just one concern - does it work with a net6.0 app even though we haven't added net6 support yet?

zhenlan: I believe it falls back to the net5.0 dependencies. It would be the same dependency for net6.0 anyway.

net5.0
Microsoft.Azure.Functions.Worker (>= 1.6.0)
Microsoft.Extensions.Configuration.AzureAppConfiguration (>= 5.0.0-preview)

PR #637, opened 2022-05-08
Update Event Hubs example to use the new ProcessPushNotification API

zhenlan:


examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
```diff
@@ -74,9 +76,14 @@ private Task ProcessEventHandler(ProcessEventArgs eventArgs)
         {
             _logger.LogInformation("EventHub update received. Triggering cache invalidation.");
 
-            //
-            // Set the cached value for key-values registered for refresh as dirty.
-            _configurationRefresher?.SetDirty();
+            // Build EventGridEvent from notification message
+            EventGridEvent eventGridEvent = EventGridEvent.Parse(BinaryData.FromBytes(eventArgs.Data.EventBody));
```
zhenlan, line 80: @MBSolomon can you please confirm this is the right code to convert Event Hubs event to Event Grid event?

MBSolomon: Here do we need to validate if EventArgs.Data is null? In the documentation it is expected to be null if the receive call times out. So this may cause issues if we try and pull the EventBody from a null value. https://docs.microsoft.com/en-us/dotnet/api/azure.messaging.eventhubs.processor.processeventargs?msclkid=bbd79e46cfb011ecba338c0ae577e165&view=azure-dotnet

avanigupta, line 80: `eventArgs.Data.EventBody` is already `BinaryData` type. `BinaryData.FromBytes` is not required.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/WebDemoWithEventHub.csproj
```diff
@@ -1,12 +1,12 @@
...
-    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.AppConfiguration.AspNetCore" Version="4.0.0-preview" />
-    <PackageReference Include="Azure.Identity" Version="1.1.1" />
-    <PackageReference Include="Azure.Messaging.EventHubs" Version="5.0.1" />
-    <PackageReference Include="Azure.Messaging.EventHubs.Processor" Version="5.0.1" />
+    <PackageReference Include="Microsoft.Azure.AppConfiguration.AspNetCore" Version="5.0.0" />
+    <PackageReference Include="Azure.Identity" Version="1.6.0" />
```
avanigupta, line 7: There is an open issue with Azure.Identity v1.5.0 and higher: https://github.com/Azure/azure-sdk-for-net/issues/24767 
I dont know if we want to stick to 1.4.1 for this example or upgrade.

zhenlan: Rolled back to 1.4.1. Thanks.

examples/DotNetCore/WebDemoWithEventHub/WebDemoWithEventHub/EventHubService.cs
zhenlan: Updated. Thanks!

zhenlan: Updated. Thanks!

avanigupta (approved):

Issue #315, opened 2020-04-14
Best Practice for handling local settings while development


LaurinSt:
Hi there

I was searching through your documentation, which has many very useful inputs, thank you for this.
I am missing one point:
What is the best practice you propose for local development machine settings? It's clear to me that in Azure App Configuration Service you can have labels for different environments. But if you have 5 developers on a project you probably want all of them to add there local dev machine settings with labels. So is there any best practice for this? Or is it just help yourself and add a fallback to local appsettings with a preprocessor constant like #if debug?
Thanks for you help

drago-draganov:
Hello @LaurinSt 

Suggestions may vary depending on your local dev environment. Is this .NET Core application or something else?

LaurinSt:
Most of the times they are .net core asp.net web api projects. But how would it differ between console project and asp.net project?

drago-draganov:
These all fall into the same category of .NET apps. I asked because we could've been talking about Java or something else.
I will assume here that you are using AzureAppConfiguration Provider as well. 

If you need to have different centralized dev profiles and share between the team, then using labels is a good start. 

If you want to allow devs to play locally with some settings (for debugging purpose or whatnot), then you can layer AzureAppConfiguration and local json file (ex. appsettings.dev.json) on top. .NET Core Configuration allows for multiple config providers. 
So use AzureAppConfiguration for all centralized settings and local appsettings.dev.json for devs to change stuff locally. Since appsetting.dev.json is after AddAzureAppConfiguration, it should override an existing centralized setting, as well as provide new as appropriate.

Perhaps something like this:

```cs
            var config = new ConfigurationBuilder()
                .AddAzureAppConfiguration(...)
                .AddJsonFile("appsettings.dev.json");
                .Build();
```

LaurinSt:
Thanks. That's a nice way I was not aware of.

iqoOopi:
@drago-draganov 
Hi drago,

We kind of want the same thing for our .net framework app. When debugging/running locally we don't want to call app configuration at all to save some cost. 
For your code:
`
var config = new ConfigurationBuilder()
                .AddAzureAppConfiguration(...)
                .AddJsonFile("appsettings.dev.json");
                .Build();`

I believe it will still make a call to the Azure app configuration then the local file will override them. Wondering how can we don't call it when local debugging.

Thanks

drago-draganov:
@iqoOopi,

```cs
var configBuilder = new ConfigurationBuilder();

//
[...]

Issue #178, opened 2019-11-01
Allow Geo-replication for App Configuration.  


msajid:
Allow Geo-replication to achieve high availability and performance benefits. Currently there is no option other than creating multiple instances of the service and replicating manually using import/export feature.

zhenlan:
Thanks for the suggestion @msajid. Yes, this is definitely something we would love to support in the future.

drago-draganov:
Please check [Geo-replication overview](https://docs.microsoft.com/en-us/azure/azure-app-configuration/concept-geo-replication)

zhenlan:
The geo-replication feature reached GA today. See announcement https://github.com/Azure/AppConfiguration-Announcements/issues/29.

PR #697, opened 2022-11-29
Add release notes for 5.2.0

amerjusupovic:


releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
```diff
@@ -1,6 +1,13 @@
 ## Microsoft.Extensions.Configuration.AzureAppConfiguration
 ### [Package (NuGet)](https://www.nuget.org/packages/Microsoft.Extensions.Configuration.AzureAppConfiguration)
 
+### 5.2.0 - November 29, 2022
```
avanigupta, line 4: These notes are describing the changes that exist in the new package vs the previous package. So our release notes should list all the differences between 5.2.0 and 5.3.0-preview.
- Removed the following new API for the App Configuration geo-replication support. [#178](https://github.com/Azure/AppConfiguration/issues/178)
   ```cs
   public AzureAppConfigurationOptions Connect(IEnumerable<Uri> endpoints, TokenCredential credential)
   ```
- Removed support for Feature Management V2 schema introduced in Microsoft.FeatureManagement 3.0.0-preview library. [#315](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/315)

avanigupta (approved):

releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
```diff
@@ -1,6 +1,18 @@
...
+* Added support for .NET 7 as a target framework. [#366](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/366)
+* Fixed a bug where passing an empty string for the label filter to `AzureAppConfigurationOptions.Select` would select key-values with all labels. [#311](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/311)
+* Fixed a bug where calls to `AzureAppConfigurationOptions.Select` would be incorrectly ignored if the key-values were already selected by a previous call. [#349](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/349)
+* Fixed a bug where `AddAzureAppConfiguration` could throw an exception on invalid feature flags even with the `optional` parameter set to `true`. [#348](https://github.com/Azure/AppConfiguration-DotnetProvider/issues/348)
+* Upgraded the package references `Azure.Security.KeyVault.Secrets` to `4.3.0` and `System.Text.Json` to `4.7.2`. [#351](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/351)
+* Removed the following new API for the App Configuration geo-replication support. [#178](https://github.com/Azure/AppConfiguration/issues/178)
```
zhenlan, line 10: While this is true compared with the last preview release, I'm a little concerned about how this would be perceived by customers :). We are not implying we are removing the geo-replication support, right?

It's really about what it compares to. The geo-replication API was never made in a stable release. I think we can drop it from the release nodes. This applies to the v2 schema too.

We didn't follow it strictly in the past, but it's not too late to start it now. For a stable version release, we should include all items that are new compared to the previous stable version, even if some of them have been mentioned in a previous preview release. It's good to call them out if a feature in a preview made it into a stable release.

PR #720, opened 2023-02-17
ConfigurationStoreBackup sample - Upgraded to .NET 6 and fixed the KeyLabel comparer issue

calloncampbell:
This addresses an issue with the sample code from this scenario:
https://learn.microsoft.com/en-us/azure/azure-app-configuration/howto-backup-config-store

Fixed an issue where the KeyLabel Equals override was not working with the current behaviour in Azure. When an app config key was being updated (with or without a Label), it was never matched against the keys found in the app config store. 

Also upgraded to .NET 6 and changed the NuGet package references to work with Azure Functions runtime v4.

Deployed updated code to Azure portal and it resolved the issue. Tested with the following scenarios:
- Added a new key to the primary store and the key was synchronized over to the secondary store (added).
- Updated existing key in the primary store and the key was synchronized over to the secondary store (updated).
- Deleted an existing key from the primary store and the key was synchronized over to the secondary store (deleted).
- Tested the above for keys with and without Labels.

### DIFF
--- a/examples/ConfigurationStoreBackup/ConfigurationStoreBackup.csproj
+++ b/examples/ConfigurationStoreBackup/ConfigurationStoreBackup.csproj
@@ -1,13 +1,13 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
-    <AzureFunctionsVersion>v3</AzureFunctionsVersion>
+    <TargetFramework>net6.0</TargetFramework>
+    <AzureFunctionsVersion>v4</AzureFunctionsVersion>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Azure.Data.AppConfiguration" Version="1.0.0" />
-    <PackageReference Include="Azure.Identity" Version="1.1.1" />
-    <PackageReference Include="Azure.Storage.Queues" Version="12.11.0" />
-    <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="3.0.6" />
+	<PackageReference Include="Microsoft.Extensions.Configuration.AzureAppConfiguration" Version="5.2.0" />
+    <PackageReference Include="Azure.Identity" Version="1.8.2" />
+    <PackageReference Include="Azure.Storage.Queues" Version="12.12.0" />
+	<PackageReference Include="Microsoft.NET.Sdk.Functions" Version="4.1.3" />
   </ItemGroup>
   <ItemGroup>
     <None Update="host.json">
--- a/examples/ConfigurationStoreBackup/ConfigurationStoreBackup.csproj
+++ b/examples/ConfigurationStoreBackup/ConfigurationStoreBackup.csproj
@@ -4,10 +4,10 @@
     <AzureFunctionsVersion>v4</AzureFunctionsVersion>
   </PropertyGroup>
   <ItemGroup>
-	<PackageReference Include="Microsoft.Extensions.Configuration.AzureAppConfiguration" Version="5.2.0" />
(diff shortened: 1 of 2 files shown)


examples/ConfigurationStoreBackup/KeyLabel.cs
```diff
@@ -10,17 +10,7 @@ internal struct KeyLabel
...
-
-        public override bool Equals(Object obj)
-        {
-            if (obj is KeyLabel keyLabel)
-            {
-                return string.Equals(this.Key, keyLabel.Key, StringComparison.Ordinal)
-                    && string.Equals(this.Label, keyLabel.Label, StringComparison.Ordinal);
-            }
-            return false;
+            Label = label ?? string.Empty;
```
avanigupta, line 13: We need an equality overload because we need to match both key and label. Key+Label is the unique identifier for any key-value in AppConfig. We can change it to case-insensitive comparison though.

examples/ConfigurationStoreBackup/ConfigurationStoreBackup.csproj
```diff
@@ -1,13 +1,13 @@
...
-    <AzureFunctionsVersion>v3</AzureFunctionsVersion>
+    <TargetFramework>net6.0</TargetFramework>
+    <AzureFunctionsVersion>v4</AzureFunctionsVersion>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Azure.Data.AppConfiguration" Version="1.0.0" />
-    <PackageReference Include="Azure.Identity" Version="1.1.1" />
-    <PackageReference Include="Azure.Storage.Queues" Version="12.11.0" />
-    <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="3.0.6" />
+	<PackageReference Include="Microsoft.Extensions.Configuration.AzureAppConfiguration" Version="5.2.0" />
```
avanigupta, line 7: This is library is not really used for backup. We just need Azure.Data.AppConfiguration. We can update that to the latest stable version.

examples/ConfigurationStoreBackup/ConfigurationStoreBackup.csproj
```diff
@@ -1,13 +1,13 @@
...
+    <PackageReference Include="Azure.Identity" Version="1.8.2" />
+    <PackageReference Include="Azure.Storage.Queues" Version="12.12.0" />
+	<PackageReference Include="Microsoft.NET.Sdk.Functions" Version="4.1.3" />
```
avanigupta, line 10: Nit: extra space

examples/ConfigurationStoreBackup/KeyLabel.cs
zhenlan: > We can change it to case-insensitive comparison though.

Why is it?

calloncampbell: So I tried that but the override was completely useless when deployed as is to Azure. When a config key with no label came through it was not being matched. Removing the override and setting a default on a null string was what fixed it.

examples/ConfigurationStoreBackup/ConfigurationStoreBackup.csproj
calloncampbell: Fixed

examples/ConfigurationStoreBackup/KeyLabel.cs
avanigupta: @zhenlan You're right, its not case insensitive. I remembered incorrectly. 
@calloncampbell thanks for catching this! We can leave the override and just assign empty string if label is null in the KeyLabel constructor (like you already have here). That should fix the issue.

calloncampbell:
Ok I'll add back the override and leave in the default. Glad I could help.

Callon
________________________________

avanigupta (approved):

PR #728, opened 2023-03-16
Python Sample Updates

mrm9084:
Remove Color Option
Added ConfigParser Example

### DIFF
--- a/examples/Python/python-django-webapp-sample/README.md
+++ b/examples/Python/python-django-webapp-sample/README.md
@@ -21,7 +21,6 @@ az appconfig create --name <config-store-name> --resource-group <resource-group-
 ```Powershell
 az appconfig kv set --name <config-store-name> --key testapp_settings_message --value "Hello from Azure App Configuration"
 az appconfig kv set --name <config-store-name> --key testapp_settings_font_size --value "30px"
-az appconfig kv set --name <config-store-name> --key testapp_settings_color --value "yellow"
 ```
 
 ### Create Key Vault Reference
--- a/examples/Python/python-django-webapp-sample/README.md
+++ b/examples/Python/python-django-webapp-sample/README.md
@@ -21,6 +21,7 @@ az appconfig create --name <config-store-name> --resource-group <resource-group-
 ```Powershell
 az appconfig kv set --name <config-store-name> --key testapp_settings_message --value "Hello from Azure App Configuration"
 az appconfig kv set --name <config-store-name> --key testapp_settings_font_size --value "30px"
+az appconfig kv set --name <config-store-name> --key testapp_settings_color --value "azure"
 ```
 
 ### Create Key Vault Reference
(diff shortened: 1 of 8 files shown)


examples/Python/python-django-webapp-sample/README.md
```diff
@@ -21,7 +21,6 @@ az appconfig create --name <config-store-name> --resource-group <resource-group-
 ```Powershell
 az appconfig kv set --name <config-store-name> --key testapp_settings_message --value "Hello from Azure App Configuration"
 az appconfig kv set --name <config-store-name> --key testapp_settings_font_size --value "30px"
-az appconfig kv set --name <config-store-name> --key testapp_settings_color --value "yellow"
```
zhenlan, line 24: Nothing wrong with it. Just be accessibility-aware if using it for demos. For example, our Azure portal is accessibility compliant, but it doesn't mean it will be colorless.

mrm9084: These are usually the first thing we bring up to demo. Also the yellow was a rather bright one… The text should be good enough.

examples/Python/python-django-webapp-sample/quickstartproject/settings.py
```diff
@@ -12,9 +12,15 @@
...
 import os
 from pathlib import Path
+import configparser
 from azure.appconfiguration.provider import load, AzureAppConfigurationKeyVaultOptions, SettingSelector
 from azure.identity import DefaultAzureCredential
 
+c_parser = configparser.ConfigParser()
+c_parser.read('static/config.ini')
+
+default = c_parser['DEFAULT']
```
zhenlan, line 22: I will stick with the `config` variable name but add a comment explaining this is loading the default configuration from an ini file.

examples/Python/python-django-webapp-sample/quickstartproject/settings.py
```diff
@@ -31,16 +37,17 @@
               selects=[selects, selects_secret],
               trim_prefixes=["testapp_settings_"])
```
zhenlan, line 38: To not conflict with the `config` variable name, we can change this to `azure_appconfiguration`.

examples/Python/python-django-webapp-sample/quickstartproject/settings.py
```diff
@@ -31,16 +37,17 @@
...
 
+default.update(config)
```
zhenlan, line 40: Add a comment explaining that the configuration loaded from appconfig will override the default configuration for overlapping keys.

mrm9084: Updated with
```python
# Updates the config object with the app configuration key-values and resolved key vault reference values.
# This will override any values in the config object with the same key.
```

examples/Python/python-django-webapp-sample/README.md
zhenlan: We have other samples that use a color background too. It's good to keep them consistent. We can change the value to something less bright if preferred, for example, azure :)

mrm9084: Added back as `Azure`

zhenlan (approved):

PR #740, opened 2023-04-12
Add Map API example with Table Storage

amerjusupovic:


amerjusupovic:
Open to suggestions for the example name other than "ConsoleApplicationWithMap"

examples/DotNetCore/ConsoleApplicationWithMap/ConsoleApplicationWithMap/appsettings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "ConnectionString": ""
```
zhenlan, line 2: Never suggest people put a connection string in a config file.

examples/DotNetCore/ConsoleApplicationWithMap/ConsoleApplicationWithMap/Program.cs
```diff
@@ -0,0 +1,158 @@
...
+                                if (blobContent != null)
+                                {
+                                    newSettingValue = $"[{string.Join(", ", blobContent.Data)}]";
+                                }
+                                else
+                                {
+                                    newSettingValue = "Error parsing Blob content.";
+                                }
+
+                                setting = new ConfigurationSetting(setting.Key, newSettingValue, setting.Label, setting.ETag);
```
avanigupta, line 119: If we dont rely on the etag of mapped ConfigurationSetting, its better to leave it out of the sample.

examples/DotNetCore/ConsoleApplicationWithMap/ConsoleApplicationWithMap/Program.cs
```diff
@@ -0,0 +1,158 @@
...
+                        .Map(async (setting) =>
+                        {
+                            if (setting.ContentType.Equals("application/storage.blob"))
+                            {
+                                MyBlobContent? blobContent = await ReadBlobContentAsync(new Uri(setting.Value));
+                                string newSettingValue = "";
+
+                                if (blobContent != null)
+                                {
+                                    newSettingValue = $"[{string.Join(", ", blobContent.Data)}]";
```
avanigupta, line 112: What is the motivation behind converting blob data into a value that looks like a list representation?

amerjusupovic: The example from the customer issue converted the blob with text in json format to a list, so I was copying that assuming it was significant somehow. I can also just directly copy the file as a string to the setting value.

avanigupta: If we want to show the full binding example, we could change the content type of mapped setting to application/json. That way, our JSON adapter will process the mapped list value and customers can bind the final IConfiguration value directly to a List object.

amerjusupovic: That sounds like a good idea, I'll test that out

amerjusupovic: Seems to work, I've pushed some changes with revision but I think I'm still unsure on how the display should look

examples/DotNetCore/ConsoleAppWithBlobStorageReferences/Program.cs
```diff
@@ -0,0 +1,139 @@
...
+            StringBuilder sb = new StringBuilder();
+
+            do
+            {
+                sb.Clear();
+
+                // Trigger and wait for an async refresh for registered configuration settings
+                await _refresher.TryRefreshAsync();
+
+                IEnumerable<string?> blobData = Configuration.GetSection("BlobUri:Data").AsEnumerable().Where(kv => !string.IsNullOrEmpty(kv.Value)).Select(kv => $"{kv.Key} = {kv.Value}");
```
avanigupta, line 123: You could directly bind configuration to MyBlobContent object.

```cs
var blobContent = new MyBlobContent();
Configuration.GetSection("BlobUri").Bind(blobContent);

sb.AppendLine($"Blob content: \n{string.Join(", ", blobContent.Data)}");

```

examples/DotNetCore/ConsoleAppWithBlobStorageReferences/Program.cs
```diff
@@ -0,0 +1,139 @@
...
+            builder.AddAzureAppConfiguration(options =>
+            {
+                options.Connect(configuration["ConnectionString"])
+                        // key: BlobUri
+                        // value: https://{account_name}.blob.core.windows.net/{container_name}/{blob_name}
+                        .Select("BlobUri")
+                        .ConfigureRefresh(refresh =>
+                        {
+                            // Changes to BlobSentinel will refresh all key-values, including BlobUri
+                            refresh.Register("BlobSentinel", true);
```
avanigupta, line 89: I feel like adding refresh in this example is making it more complicated and taking the focus away from Map API. Can you remove refresh and simplify the example? For example, we dont need a separate `Run` method and we can directly display blob content in main method.

examples/DotNetCore/ConsoleAppWithBlobStorageReferences/ConsoleAppWithBlobStorageReferences.csproj
```diff
@@ -0,0 +1,24 @@
...
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Azure.Identity" Version="1.8.2" />
+    <PackageReference Include="Azure.Storage.Blobs" Version="12.15.1" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.AzureAppConfiguration" Version="6.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="7.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="7.0.0" />
```
avanigupta, line 15: Are these two package references needed?

examples/DotNetCore/ConsoleAppWithBlobStorageReferences/Program.cs
amerjusupovic: Yea it does look a bit complicated, I just assumed that the customer use case would most of the time involve a refresh. I can remove it and see how it looks.

amerjusupovic:
I serialize the data to a json similar to the previous blob example, and it also demonstrates how the internal json adapter runs after Map finishes. However, I understand that turning the data into a json and back may seem odd so I'm open to suggestions there.

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
```diff
@@ -0,0 +1,140 @@
...
+            }
+
+            // Augment the configuration builder with Azure App Configuration
+            // Pull the connection string from an environment variable
+            builder.AddAzureAppConfiguration(options =>
+            {
+                options.Connect(configuration["ConnectionString"])
+                        // key: TableUri
+                        // value: https://{account_name}.table.core.windows.net/{table_name}
+                        .Select("TableUri")
```
zhenlan, line 122: You will want to load all configurations for your application and the inventory is just one of key-values. For example, I may have a key "Myapp:Inventory" and its value just happen to be a reference to a storage table.

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
```diff
@@ -0,0 +1,140 @@
...
+            Configuration.GetSection("TableUri").Bind(tableContent);
+
+            sb.AppendLine("Your store inventory:\n");
+
+            var tableFormat = "{0,-20} {1,-10} {2,-10}";
+
+            sb.AppendFormat(tableFormat, "Product Name", "Quantity", "On Sale");
+            sb.AppendLine();
+
+            foreach (Product product in tableContent)
```
zhenlan, line 49: Nothing wrong with this, but it kind of counters the purpose of showing off the configuration. I would just show the whole inventory based on the configuration (everything under the "Inventory" node).

amerjusupovic: Not sure I understand, are you saying instead of how I bound it to `tableContent`, just to print the configuration section to the console?

zhenlan: Right. While the data is saved as a table in storage, the whole idea is to make it available as a configuration. Your application should be dealing with configuration, not a table. I can provide that same data in `appsettings.json` like below and there should be no difference to the core business of an application, which is displaying the data in your example.

```json
"MyShop" : {
    "Inventory" : [
        {
            "ProductName": "Foo",
            "Quantity": 98,
            "OnSale" : true
        },
        {
            "ProductName": "bar",
            "Quantity": 1,
            "OnSale" : false
        }
    ]
}
```

amerjusupovic: From what I've tested, even with binding the data to a list of products this kind of setup from `appsettings.json` should yield the same result as long as the keys are the same. From what I can tell the inventory value is a list so if I just print the value it's null unless I bind it this way, so let me know if this works. I've added a similar example to the one you commented to `appsettings.json` as well.

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
```diff
@@ -0,0 +1,140 @@
...
+            StringBuilder sb = new StringBuilder();
+
+            List<Product> tableContent = new List<Product>();
+            Configuration.GetSection("TableUri").Bind(tableContent);
+
+            sb.AppendLine("Your store inventory:\n");
+
+            var tableFormat = "{0,-20} {1,-10} {2,-10}";
+
+            sb.AppendFormat(tableFormat, "Product Name", "Quantity", "On Sale");
```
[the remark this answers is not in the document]
amerjusupovic: I can see that the code got a bit more complicated, but running the program is still simple for a user so I think it should be ok. Just set 1-2 key-values.

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
```diff
@@ -0,0 +1,182 @@
...
+            Configure();
+
+            string display = string.Empty;
+
+            StringBuilder sb = BuildProductTable();
+
+            sb.AppendLine();
+            sb.AppendLine("Press any key to exit...");
+
+            display = sb.ToString();
```
avanigupta, line 43: display string is not needed, we can directly print sb.ToString()

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
```diff
@@ -0,0 +1,182 @@
...
+
+        private static StringBuilder BuildProductTable()
+        {
+            StringBuilder sb = new StringBuilder();
+
+            List<Product> tableContent = new List<Product>();
+            Configuration.GetSection("MyShop:Inventory").Bind(tableContent);
+
+            // Example value for MyShop:DisplayedColumns: Product OnSale
+            string myShopDisplayedColumns = Configuration.GetValue<string>("MyShop:DisplayedColumns");
```
avanigupta, line 60: What do you think about storing this value as a list of strings with application/json content type? 
// Example value for MyShop:DisplayedColumns = [Product, OnSale]
// Content-Type = application/json

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
avanigupta: @zhenlan I think we should still bind configuration to `List<Product>` because I assume that's how real applications would use it.

But this example can be simplified further if we dont display the config as a table. How about something like this:

```cs
foreach (Product product in tableContent)
{
    if (columnsToDisplay.Contains(nameof(product.Name)))
    {
        sb.AppendLine($"Name = {product.Name}");
    }
   
    if (columnsToDisplay.Contains(nameof(product.Quantity)))
    {
        sb.AppendLine($"Quantity = {product.Quantity.ToString()}");
    }

    if (columnsToDisplay.Contains(nameof(product.OnSale)))
    {
        sb.AppendLine($"OnSale = {product.OnSale.ToString()}");
    }

    sb.AppendLine();
}

```

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
```diff
@@ -0,0 +1,182 @@
...
+
+            await tableClient.UpsertEntityAsync(prod1);
+            await tableClient.UpsertEntityAsync(prod2);
+
+            var products = tableClient.Query<Product>();
+
+            return JsonSerializer.Serialize(products.ToList());
+        }
+
+        private static void Configure()
```
avanigupta, line 142: Can you move this method to the top, right after main method? It just highlights our product logic before table manipulations.

examples/DotNetCore/ConsoleAppWithTableStorageReference/appsettings.json
```diff
@@ -0,0 +1,16 @@
+{
```
avanigupta, line 1: If we're keeping sample values here, can you add a comment explaining this is the content from Table storage? Also, we should add `MyShop:DisplayedColumns` here too.

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
```diff
@@ -0,0 +1,176 @@
...
+        private static StringBuilder BuildProductTable()
+        {
+            StringBuilder sb = new StringBuilder();
+
+            List<Product> tableContent = new List<Product>();
+            Configuration.GetSection("MyShop:Inventory").Bind(tableContent);
+
+            // Example value for MyShop:DisplayedColumns = [Name, OnSale]
+            // Content-Type = application/json
+            string[] columnsToDisplay = Configuration.GetValue<string[]>("MyShop:DisplayedColumns");
```
avanigupta, line 99: Unless there's a specific reason for using array here, can we use List<string>?

examples/DotNetCore/ConsoleAppWithTableStorageReference/appsettings.json
```diff
@@ -0,0 +1,18 @@
+// Example shop inventory data. This data will be used if the 'MyShop:Inventory' key is not set.
```
avanigupta, line 1:
```suggestion
// Example shop inventory data. This data will be used if the 'MyShop:Inventory' key is not set appropriately in App Configuration.
```

avanigupta (approved):
I just think that the tabular display is making the code more complicated than it needs to be. Thoughts @zhenlan?

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
```diff
@@ -0,0 +1,161 @@
...
+
+            // Augment the configuration builder with Azure App Configuration
+            // Pull the connection string from an environment variable
+            builder.AddAzureAppConfiguration(options =>
+            {
+                options.Connect(configuration["ConnectionString"])
+                        .Select("*")
+                        .Map(async (setting) =>
+                        {
+                            if (setting.ContentType.Equals("application/storage.table"))
```
jimmyca15, line 74: application/x.foobar

See [here](https://datatracker.ietf.org/doc/html/rfc6838#section-3.4)

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
```diff
@@ -0,0 +1,161 @@
...
+            var prod2 = new Product()
+            {
+                RowKey = "68719518390",
+                PartitionKey = "gear-surf-surfboards",
+                Name = "Sand Surfboard",
+                Quantity = 5,
+                OnSale = false
+            };
+
+            await tableClient.UpsertEntityAsync(prod1);
```
jimmyca15, line 153: This is performing a Write? Doesn't match the method name.

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
```diff
@@ -0,0 +1,161 @@
...
+            builder.AddAzureAppConfiguration(options =>
+            {
+                options.Connect(configuration["ConnectionString"])
+                        .Select("*")
+                        .Map(async (setting) =>
+                        {
+                            if (setting.ContentType.Equals("application/storage.table"))
+                            {
+                                // Example value for key "MyShop:Inventory" in App Configuration: https://{account_name}.table.core.windows.net/{table_name}
+                                string tableContent = await ReadTableContentAsync(new Uri(setting.Value));
```
jimmyca15, line 77: It works but it seems like an odd use of table to me. This is more like blob usage. So my suggestion would either be to update to use blob storage, or update the reference to include table specific identifiers like partition and row key.

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
amerjusupovic: Yea wanted to have some data in the table by default, I could change the method name to reflect that.

jimmyca15: How about writing the data in the beginning of the application.

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
jimmyca15: On second thought. I think this is fine.

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
jimmyca15: The referenced items are specifically processed as `Product`. The content type should make that connection.

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
jimmyca15, line 77:
```suggestion
                                IEnumerable<Product> products = await ReadProducts(new Uri(setting.Value));
                                
                                 setting = new ConfigurationSetting(
                                     setting.Key,
                                     JsonSerializer.Serialize(products),
                                     setting.Label);
```

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
```diff
@@ -0,0 +1,161 @@
...
+                    sb.AppendLine($"OnSale = {product.OnSale.ToString()}");
+                }
+
+                sb.AppendLine();
+            }
+
+            return sb;
+        }
+
+        private static async Task<string> ReadTableContentAsync(Uri tableUri)
```
jimmyca15, line 125: All async methods should take a cancellation token.

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
```diff
@@ -0,0 +1,161 @@
...
+                            return setting;
+                        });
+            });
+
+            Configuration = builder.Build();
+        }
+
+        private static StringBuilder BuildProductTable()
+        {
+            StringBuilder sb = new StringBuilder();
```
jimmyca15, line 92: Please use consistent line spacing throughout the entire file.

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
amerjusupovic: like application/x.table?

jimmyca15: sure

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
amerjusupovic: I'm not sure where that would be, since we actually get the Uri when it's pulled from app config.

jimmyca15: Good point. Needs some thought, but writing it in the exact same code path as reading it is weird.

jimmyca15: Or, don't worry about writing data in the example. I don't believe our other examples do such a thing.

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
```diff
@@ -0,0 +1,161 @@
...
+using Azure.Data.AppConfiguration;
+using Azure.Data.Tables;
+using Azure.Identity;
+using System.Text;
+using System.Text.Json;
+
+namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Examples.ConsoleAppWithTableStorageReference
+{
+    // C# record type for items in the table
+    public record Product : ITableEntity
```
jimmyca15, line 11: Deserves it's own file.

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
jimmyca15: `application/x.example.table.product

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
```diff
@@ -0,0 +1,127 @@
...
+{
+    class Program
+    {
+        static IConfiguration Configuration { get; set; }
+
+        static void Main(string[] args)
+        {
+            Configure();
+
+            StringBuilder sb = BuildProductTable();
```
jimmyca15, line 17:
```suggestion
            var products = new List<Product>();

            Configuration.GetSection("MyShop:Inventory").Bind(products)
            
            Console.Clear();

            Console.WriteLine(FormatProducts(products));
            
            Console.WriteLine("\nPress any key to exit...");

            // Finish on key press
            Console.ReadKey();
```

jimmyca15: Can `Configuration.GetSection("MyShop:Inventory").Bind(products)` become `Configuration.GetSection("MyShop:Inventory").Get<List<Product>>()`? I seem to remember such an API.

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
```diff
@@ -0,0 +1,127 @@
...
+                return;
+            }
+
+            // Augment the configuration builder with Azure App Configuration
+            // Pull the connection string from an environment variable
+            builder.AddAzureAppConfiguration(options =>
+            {
+                options.Connect(configuration["ConnectionString"])
+                        .Select("*")
+                        .Map((setting) =>
```
jimmyca15, line 55: I would suggest having the whole map handler as the private method.

jimmyca15:
```
.Map((setting) =>
{
    switch (setting.ContentType)
    {
        case "application/x.example.tablereference.product"

            return MapProductTableReference(setting);

        default:

            return setting;
    }
}
```

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
```diff
@@ -0,0 +1,127 @@
...
+        private static IEnumerable<Product> ReadProducts(Uri tableUri)
+        {
+            string[] pathSegments = tableUri.Segments;
+
+            // The last segment in the path should be the table name
+            string tableName = pathSegments[pathSegments.Length - 1];
+
+            var tableClient = new TableClient(tableUri, tableName, new DefaultAzureCredential());
+
+            return tableClient.Query<Product>();
```
jimmyca15, line 124: Querying a table isn't async ?

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
jimmyca15, line 55: I'm suprised this isn't async given that we're reading from db.

jimmyca15:
This example seems like it should have some simple readme of what data is expected to be in app configuration as well as in table storage.

examples/DotNetCore/ConsoleAppWithTableStorageReference/Program.cs
amerjusupovic: I noticed that in the [documentation](https://learn.microsoft.com/en-us/azure/cosmos-db/table/quickstart-dotnet?toc=https%3A%2F%2Flearn.microsoft.com%2Fen-us%2Fazure%2Fstorage%2Ftables%2Ftoc.json&bc=https%3A%2F%2Flearn.microsoft.com%2Fen-us%2Fazure%2Fbread%2Ftoc.json&tabs=azure-cli%2Cwindows#query-items) they use `Query` which isn't async, but not sure why. It looks like there's also a `QueryAsync` that we could use though.

jimmyca15: Please do.

amerjusupovic: I misunderstood what `QueryAsync` was since it returns an `AsyncPageable` and not a task, which means I can't await it directly.

jimmyca15: AsyncPageable can be converted to IEnumerable in an async manner though.

amerjusupovic: I think I was missing the point of making it async, fixed it.

jimmyca15 (approved):

PR #758, opened 2023-05-15
Add requirement type to schema and bumps version from 1.0.0 to 1.1.0

rossgrambo:


docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json
```diff
@@ -56,6 +56,16 @@
...
       "properties": {
+        "requirement_type": {
+          "$id": "#/properties/conditions/properties/requirement_type",
+          "type": "string",
+          "title": "Requirement Type",
+          "description": "Determines whether any or all registered feature filters must be enabled for the feature to be considered enabled.",
+          "examples": [
+            "Any",
+            "All"
+          ]
```
zhenlan, line 67: What's the default?

rossgrambo: Default functionality is Any. However here, I was treating the default as excluding the field (As FeatureManagement decides what the "default" is), but perhaps default means "when the field is included, what is the default". 

What do you think?

docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json
```diff
@@ -56,6 +56,16 @@
       "description": "The declaration of conditions used to dynamically enable features.",
       "required": [],
       "properties": {
+        "requirement_type": {
+          "$id": "#/properties/conditions/properties/requirement_type",
+          "type": "string",
+          "title": "Requirement Type",
+          "description": "Determines whether any or all registered feature filters must be enabled for the feature to be considered enabled.",
+          "examples": [
```
jimmyca15, line 64: How about to explicitly call out enum usage.

```
          "enum": ["Any", "All"],
```

docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json
```diff
@@ -56,6 +56,16 @@
       "description": "The declaration of conditions used to dynamically enable features.",
       "required": [],
       "properties": {
+        "requirement_type": {
+          "$id": "#/properties/conditions/properties/requirement_type",
+          "type": "string",
+          "title": "Requirement Type",
+          "description": "Determines whether any or all registered feature filters must be enabled for the feature to be considered enabled.",
```
jimmyca15, line 63: > feature filters

To be consistent with the schema, client filters may be preferred.

docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json
jimmyca15: We can include a [default annotation](https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.7.7.1.1) just for informational purposes.

docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json
```diff
@@ -56,6 +56,14 @@
       "description": "The declaration of conditions used to dynamically enable features.",
       "required": [],
       "properties": {
+        "requirement_type": {
+          "$id": "#/properties/conditions/properties/requirement_type",
+          "type": "string",
+          "title": "Requirement Type",
+          "description": "Determines whether any or all registered client filters must be enabled for the feature to be considered enabled.",
+          "enum": [ "Any", "All" ],
```
jimmyca15, line 64: I see a few other small arrays, none inlined though.

jimmyca15 (approved):

docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json
jimmyca15: I realize that I inlined it in my suggestion, whoops. Would still recommend not to do so since it's not done elsewhere in the file.

zhenlan (approved):

Issue #760, opened 2023-05-17
Azure App Configuration Push Task@5: Undocumented breaking change with feature flags


supereddie:
With version 5 of AzureAppConfigurationPush it is no longer possible to push settings with the feature flag prefix, resulting in the following error: `Key cannot start with the reserved prefix for feature flags.`

The version 4 of the task works without issue:
```yaml
- task: AzureAppConfigurationPush@4
  displayName: 'Feature flags - Development'
  inputs:
    azureSubscription: '${{ parameters.serviceConnection }}'
    AppConfigurationEndpoint: '$(appConfigurationEndpoint)'
    ConfigurationFile: '$(Pipeline.Workspace)/${{ parameters.artifactName }}/feature-flags-test.json'
    Separator: '-'
    Prefix: '.appconfig.featureflag/'
    Depth: '1'
    Label: 'Development'
    ContentType: 'application/vnd.microsoft.appconfig.ff+json;charset=utf-8'
    Strict: true
```

Contents of .json file:
```json
{
  "MyFlagName": {
    "id": "MyFlagName",
    "description": "",
    "enabled": true,
    "conditions": {
      "client_filters": []
    }
  }
}
```

I cannot find this breaking change in the release notes of the task, so I don't know if this change was intentional or not.

MaryanneNjeri:
Hello @supereddie, seems we'll need to update the release notes to document it as a breaking change. Apologies for any inconvenience caused. I'll send an update for that.

RichardChen820:
Since your test json is not using the correct schema that feature flag content type required, hence the `MyFlagName` would be treated as a normal key-value, turns out you were trying to add a feature flag prefix to a normal key, that's the reason it throws the error.

MaryanneNjeri:
PR https://github.com/Azure/AppConfiguration/pull/762 has been merged in

PR #762, opened 2023-05-24
Updated to document breaking changes

MaryanneNjeri:
This PR updates the AzureDevOpsPushPipelineExtenstion.md and documents importing feature flag breaking change 
GitHub issue #760

### DIFF
--- a/releaseNotes/AzureDevOpsPipelineExtension.md
+++ b/releaseNotes/AzureDevOpsPipelineExtension.md
@@ -5,7 +5,7 @@ The Azure App Configuration extension in Azure DevOps pipeline has been publishe
 ![sample](pictures/AzureDevOpsExtensionVersionSample.PNG)
 
 ### v6.0.0 - February, 02 2023
-**Breaking changes**:
+**Breaking changes**
   - Updated the task to require Node.js 16. It previously required 10.
   - Updated the minimum supported azure pipeline agent version to [2.206.1](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.206.1) or later. Previously it was [2.144.0](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.144.0).
 
--- a/releaseNotes/AzureDevOpsPipelineExtension.md
+++ b/releaseNotes/AzureDevOpsPipelineExtension.md
@@ -5,11 +5,12 @@ The Azure App Configuration extension in Azure DevOps pipeline has been publishe
 ![sample](pictures/AzureDevOpsExtensionVersionSample.PNG)
 
 ### v6.0.0 - February, 02 2023
-* **Breaking changes**:
+**Breaking changes**:
   - Updated the task to require Node.js 16. It previously required 10.
   - Updated the minimum supported azure pipeline agent version to [2.206.1](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.206.1) or later. Previously it was [2.144.0](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.144.0).
 
-* Fixed a bug that caused an invalid secret URL warning when resolving key vault references in sovereign clouds.
+**Bug fixes**
+  - Fixed a bug that caused an invalid secret URL warning when resolving key vault references in sovereign clouds.
 
 ### v5.2.0 - March, 02 2022
 Fixed a bug that caused an invalid_client error when using certificate-based authentication [#608](https://github.com/Azure/AppConfiguration/issues/608).
--- a/releaseNotes/AzureDevOpsPipelineExtension.md
+++ b/releaseNotes/AzureDevOpsPipelineExtension.md
@@ -5,7 +5,15 @@ The Azure App Configuration extension in Azure DevOps pipeline has been publishe
 ![sample](pictures/AzureDevOpsExtensionVersionSample.PNG)
 
 ### v6.0.0 - February, 02 2023
-* Updated the task to require Node.js 16. It previously required 10.
+* **Breaking change**:
+  With the update to the task to require Node.js 16. It previously required 10. The minimum agent version for running the task on self hosted agent changed. 
+
+    **Before**:
+        [Azure Pipeline agent version 2.144.0](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.144.0) or later.
+
+    **After**:
+        [Azure Pipeline agent version 2.206.1](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.206.1) or later.
+
 * Fixed a bug that caused an invalid secret URL warning when resolving key vault references in sovereign clouds.
 
 ### v5.2.0 - March, 02 2022
--- a/releaseNotes/AzureDevOpsPipelineExtension.md
+++ b/releaseNotes/AzureDevOpsPipelineExtension.md
@@ -5,14 +5,9 @@ The Azure App Configuration extension in Azure DevOps pipeline has been publishe
 ![sample](pictures/AzureDevOpsExtensionVersionSample.PNG)
 
 ### v6.0.0 - February, 02 2023
-* **Breaking change**:
-  With the update to the task to require Node.js 16. It previously required 10. The minimum agent version for running the task on self hosted agent changed. 
-
-    **Before**:
-      [Azure Pipeline agent version 2.144.0](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.144.0) or later.
-
-    **After**:
-      [Azure Pipeline agent version 2.206.1](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.206.1) or later.
+* **Breaking changes**:
+  - Updated the task to require Node.js 16. It previously required 10.
+  - Updated the minimum supported azure pipeline agent version to [2.206.1](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.206.1) or later. Previously it was [2.144.0](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.144.0).
 
 * Fixed a bug that caused an invalid secret URL warning when resolving key vault references in sovereign clouds.
 
--- a/releaseNotes/AzureDevOpsPipelineExtension.md
+++ b/releaseNotes/AzureDevOpsPipelineExtension.md
@@ -9,10 +9,10 @@ The Azure App Configuration extension in Azure DevOps pipeline has been publishe
   With the update to the task to require Node.js 16. It previously required 10. The minimum agent version for running the task on self hosted agent changed. 
 
     **Before**:
-        [Azure Pipeline agent version 2.144.0](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.144.0) or later.
+      [Azure Pipeline agent version 2.144.0](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.144.0) or later.
 
     **After**:
-        [Azure Pipeline agent version 2.206.1](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.206.1) or later.
+      [Azure Pipeline agent version 2.206.1](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.206.1) or later.
 
 * Fixed a bug that caused an invalid secret URL warning when resolving key vault references in sovereign clouds.
 
--- a/releaseNotes/AzureDevOpsPushPipelineExtension.md
+++ b/releaseNotes/AzureDevOpsPushPipelineExtension.md
@@ -3,7 +3,7 @@ Azure App Configuration Push extension for Azure DevOps pipeline can be installe
 ![sample](pictures/AzureDevOpsPushExtensionVersionSample.PNG)
 
 ### v5.0.0 - February, 02 2023
-**Breaking Changes**:
+**Breaking Changes**
   - Updated the task to require Node.js 16. It previously required 10.
   - Updated the minimum supported azure pipeline agent version to [2.206.1](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.206.1) or later. Previously it was [2.144.0](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.144.0).
   - Importing key vault references value as an escaped json strings is not supported, please specify the key vault reference value as a json object.
--- a/releaseNotes/AzureDevOpsPushPipelineExtension.md
+++ b/releaseNotes/AzureDevOpsPushPipelineExtension.md
@@ -6,7 +6,7 @@ Azure App Configuration Push extension for Azure DevOps pipeline can be installe
 **Breaking Changes**
   - Updated the task to require Node.js 16. It previously required 10.
   - Updated the minimum supported azure pipeline agent version to [2.206.1](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.206.1) or later. Previously it was [2.144.0](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.144.0).
-  - Importing configurations with values as escaped JSON strings and the content type application/json demonstrates an altered behavior. Please indicate the value as a JSON object.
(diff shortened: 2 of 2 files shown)


releaseNotes/AzureDevOpsPushPipelineExtension.md
```diff
@@ -3,7 +3,15 @@ Azure App Configuration Push extension for Azure DevOps pipeline can be installe
 ![sample](pictures/AzureDevOpsPushExtensionVersionSample.PNG)
 
 ### v5.0.0 - February, 02 2023
-Updated the task to require Node.js 16. It previously required 10
+* Updated the task to require Node.js 16. It previously required 10
```
zhenlan, line 6: This is a breaking change too.

releaseNotes/AzureDevOpsPushPipelineExtension.md
```diff
@@ -3,7 +3,15 @@ Azure App Configuration Push extension for Azure DevOps pipeline can be installe
...
-Updated the task to require Node.js 16. It previously required 10
+* Updated the task to require Node.js 16. It previously required 10
+
+* **Breaking change**: 
+
+    **Before**: 
+      Import feature flags to store by specifying the content type as “application/vnd.microsoft.appconfig.ff+json;charset=utf-8 ” and specifying the prefix as “.appconfig.featureflag/”.
+
+    **After**:
+      Import feature flags to store by specifying the prefix as “.appconfig.featureflag/” is not supported and it will be treated as key-value. To import feature flags please follow [feature management schema](https://github.com/microsoft/FeatureManagement-Dotnet/blob/release/v3/docs/schemas/FeatureManagement.v1.0.0.json).
```
zhenlan, line 14: Is it true? I thought it was prohibited.

releaseNotes/AzureDevOpsPushPipelineExtension.md
MaryanneNjeri: Yes, updated to document the breaking change.

releaseNotes/AzureDevOpsPushPipelineExtension.md
MaryanneNjeri: Sorry my bad, it's prohibited,  we throw an error to let user know.

releaseNotes/AzureDevOpsPushPipelineExtension.md
```diff
@@ -3,7 +3,22 @@ Azure App Configuration Push extension for Azure DevOps pipeline can be installe
...
+* **Breaking change**:
+  With the update to the task to require Node.js 16. It previously required 10. The minimum agent version for running the task on self hosted agent changed. 
+
+    **Before**:
+        [Azure Pipeline agent version 2.144.0](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.144.0) or later.
+
+    **After**:
+        [Azure Pipeline agent version 2.206.1](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.206.1) or later.
+
+* **Breaking change**:
```
RichardChen820, line 15: How about the escaped Key Vault reference json string issue that is reported by the customer? It's another break change

MaryanneNjeri: Yes, let me make the update

releaseNotes/AzureDevOpsPushPipelineExtension.md
```diff
@@ -3,7 +3,31 @@ Azure App Configuration Push extension for Azure DevOps pipeline can be installe
...
+    **Before**:
+        [Azure Pipeline agent version 2.144.0](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.144.0) or later.
+
+    **After**:
+        [Azure Pipeline agent version 2.206.1](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.206.1) or later.
+
+* **Breaking change**: 
+
+    **Before**: 
+      Import feature flags to store by specifying the content type as “application/vnd.microsoft.appconfig.ff+json;charset=utf-8 ” and specifying the prefix as “.appconfig.featureflag/”.
```
RichardChen820, line 18: Remove extra whitespace

releaseNotes/AzureDevOpsPushPipelineExtension.md
```diff
@@ -3,7 +3,30 @@ Azure App Configuration Push extension for Azure DevOps pipeline can be installe
...
+* **Breaking change**:
+  With the update to the task to require Node.js 16. It previously required 10. The minimum agent version for running the task on self hosted agent changed.
+
+    **Before**:
+      [Azure Pipeline agent version 2.144.0](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.144.0) or later.
+
+    **After**:
+      [Azure Pipeline agent version 2.206.1](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.206.1) or later.
+
+* **Breaking change**:
```
jimmyca15, line 15: I'd recommend having a single breaking changes section with bullet points below of the breaking changes.

## Breaking Changes

* Updated the task to require Node.js 16. It previously required 10.
* Updated the minimum supported agent version to 2.206.1. Previously it was 2.144.0

releaseNotes/AzureDevOpsPushPipelineExtension.md
```diff
@@ -3,7 +3,30 @@ Azure App Configuration Push extension for Azure DevOps pipeline can be installe
...
+
+    **Before**:
```
jimmyca15, line 17: The before and after is a bit hard to follow. I'd like if we can be a bit more direct on what's broken since these are often used for reference. It sounds like maybe it should be:

* Feature flag import now requires adherence to the [feature management schema](https://github.com/microsoft/FeatureManagement-Dotnet/blob/release/v3/docs/schemas/FeatureManagement.v1.0.0.json). Settings with feature flag prefix and content type will fail to import.



---

A follow up question here. Why did this break?

MaryanneNjeri: Updated to have the Breaking changes with bullet point and removed the before and after.

MaryanneNjeri: In version 5 of the Push task, we  integrated with the sync lib, which tries to have the import behavior consistent with the CLI, the CLI logs a warning when importing kvs, and key starts with feature flag prefix, https://github.com/Azure/azure-cli/blob/fde406f3aa731d33e08ef7a77404b25f10f3b559/src/azure-cli/azure/cli/command_modules/appconfig/_kv_helpers.py#L108

releaseNotes/AzureDevOpsPushPipelineExtension.md
MaryanneNjeri: Updated to have a single breaking changes section with bullet points. 
To confirm for the Pull task since we only document one breaking change ie on the Node js upgrade, we don't need to have it in bullet points and can have the before and after sections?

releaseNotes/AzureDevOpsPushPipelineExtension.md
jimmyca15: The CLI logs a warning, yet this fails outright? Doesn't seem like consistent behavior.

MaryanneNjeri: Since Sync client lib doesn't take a logger object,  all warnings are treated as errors in the Sync lib. Though am thinking maybe the Push task can catch this error and log it as a warning.

jimmyca15: The portal doesn't allow creation of feature flags as raw KVs.

Push task doesn't allow it (our current discussion)

but CLI does allow it.

I think first we need to align on what we want the behavior to be across all client side tools.

releaseNotes/AzureDevOpsPipelineExtension.md
```diff
@@ -5,7 +5,15 @@ The Azure App Configuration extension in Azure DevOps pipeline has been publishe
 ![sample](pictures/AzureDevOpsExtensionVersionSample.PNG)
 
 ### v6.0.0 - February, 02 2023
-* Updated the task to require Node.js 16. It previously required 10.
+* **Breaking change**:
+  With the update to the task to require Node.js 16. It previously required 10. The minimum agent version for running the task on self hosted agent changed.
```
jimmyca15, line 9: This needs the same update as the other file.

MaryanneNjeri: Updated and removed the before/after

releaseNotes/AzureDevOpsPushPipelineExtension.md
MaryanneNjeri: The Portal supports importing configurations from a file and specifying the prefix as ".appconfig.featureflag/" and content type "application/vnd.microsoft.appconfig.ff+json;charset=utf-8"

releaseNotes/AzureDevOpsPipelineExtension.md
```diff
@@ -5,7 +5,10 @@ The Azure App Configuration extension in Azure DevOps pipeline has been publishe
 ![sample](pictures/AzureDevOpsExtensionVersionSample.PNG)
 
 ### v6.0.0 - February, 02 2023
-* Updated the task to require Node.js 16. It previously required 10.
+* **Breaking changes**:
+  - Updated the task to require Node.js 16. It previously required 10.
+  - Updated the minimum supported azure pipeline agent version to [2.206.1](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.206.1) or later. Previously it was [2.144.0](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.144.0).
+
 * Fixed a bug that caused an invalid secret URL warning when resolving key vault references in sovereign clouds.
```
jimmyca15, line 12: Is this a breaking change? If not, a new section header is needed to separate it from **Breaking changes**, such as **Bug fixes**

releaseNotes/AzureDevOpsPipelineExtension.md
```diff
@@ -5,7 +5,10 @@ The Azure App Configuration extension in Azure DevOps pipeline has been publishe
 ![sample](pictures/AzureDevOpsExtensionVersionSample.PNG)
 
 ### v6.0.0 - February, 02 2023
-* Updated the task to require Node.js 16. It previously required 10.
+* **Breaking changes**:
```
jimmyca15, line 8: Breaking changes shouldn't be a bullet, it should be a header.

releaseNotes/AzureDevOpsPipelineExtension.md
MaryanneNjeri: Yes its not a breaking change, added new section header "Bug fixes"

releaseNotes/AzureDevOpsPipelineExtension.md
MaryanneNjeri: Fixed breaking change header

releaseNotes/AzureDevOpsPushPipelineExtension.md
```diff
@@ -3,7 +3,11 @@ Azure App Configuration Push extension for Azure DevOps pipeline can be installe
 ![sample](pictures/AzureDevOpsPushExtensionVersionSample.PNG)
 
 ### v5.0.0 - February, 02 2023
-Updated the task to require Node.js 16. It previously required 10
+**Breaking Changes**
+  - Updated the task to require Node.js 16. It previously required 10.
+  - Updated the minimum supported azure pipeline agent version to [2.206.1](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.206.1) or later. Previously it was [2.144.0](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.144.0).
+  - Importing key vault references value as an escaped json strings is not supported, please specify the key vault reference value as a json object.
```
jimmyca15, line 9: This is a change for all files using application/json content type right, not just kv set? If so, we'll want to call that out.

Also, may help to provide a snippet of what has changed.

Something like

Before when you had this

```
bla bla
```

you got this
```
bla bla bla
```

Now, the input needs to be 
```
bla bla bla
```

MaryanneNjeri: It only applies for default content profile, using application/json content type. 
How about something like :
"Importing configurations with values as escaped json strings and the content type "application/json" is no longer supported. Please specify value as json object."

**Before**
 Sample configuration file
```
{  
  "app": "{\"uri\":\"https://keyvault.vault.azure.net/secrets/secret\"}"
}
```
Results:

| Key           | Value         |
| ----------- | ----------- |
| app           |   "{"uri":"https://keyvault.vault.azure.net/secrets/secret"}"        |

**After**
The input should be

```
{  
  "app": {
      "uri": "https://keyvault.vault.azure.net/secrets/secret " 
     }
}
```

MaryanneNjeri: I was thinking since the breaking change is in relation to importing configurations, we can  show the before and after of the snipped that has changed and skip showing the results

jimmyca15: >How about something like :
"Importing configurations with values as escaped json strings and the content type "application/json" is no longer supported. Please specify value as a json object."


I would avoid the phrase "is no longer supported." It's not that we removed support but rather we changed the behavior. We now convert the values of properties in the settings to json strings.

> I was thinking since the breaking change is in relation to importing configurations, we can show the before and after of the snipped that has changed and skip showing the results

Sounds reasonable. But, I would include the content type that goes with the associated input snippet

MaryanneNjeri: > I would avoid the phrase "is no longer supported." It's not that we removed support but rather we changed the behavior. We now convert the values of properties in the settings to json strings.

Yes good point.

MaryanneNjeri: This has been updated thanks

releaseNotes/AzureDevOpsPushPipelineExtension.md
```diff
@@ -3,7 +3,43 @@ Azure App Configuration Push extension for Azure DevOps pipeline can be installe
...
+    task: AzureAppConfigurationPush@5
+    inputs:
+      azureSubscription: '<subscription>'
+      AppConfigurationEndpoint: '<store endpoint>'
+      ....
+      ContentType: 'application/vnd.microsoft.appconfig.keyvaultref+json;charset=utf-8'
+
+    ```
+
+  - Feature flag import now requires adherence to the [feature management schema](https://github.com/microsoft/FeatureManagement-Dotnet/blob/release/v3/docs/schemas/FeatureManagement.v1.0.0.json). Settings with feature flag prefix and content type will fail to import
```
bbrandt, line 42: Does specifying the feature flags as the [feature management schema](https://github.com/microsoft/FeatureManagement-Dotnet/blob/release/v3/docs/schemas/FeatureManagement.v1.0.0.json) for import then get converted to the App Config service prefix and content type for feature flags by the extension or the app config service?  So that the imported feature flags appear in the Feature Manager view alongside feature flags manually created in the portal?

releaseNotes/AzureDevOpsPushPipelineExtension.md
```diff
@@ -3,7 +3,43 @@ Azure App Configuration Push extension for Azure DevOps pipeline can be installe
 ![sample](pictures/AzureDevOpsPushExtensionVersionSample.PNG)
 
 ### v5.0.0 - February, 02 2023
-Updated the task to require Node.js 16. It previously required 10
+**Breaking Changes**
+  - Updated the task to require Node.js 16. It previously required 10.
+  - Updated the minimum supported azure pipeline agent version to [2.206.1](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.206.1) or later. Previously it was [2.144.0](https://github.com/microsoft/azure-pipelines-agent/releases/tag/v2.144.0).
+  - Importing configurations with values as escaped JSON strings and the content type `application/json`, introduces a modified behavior, the values of properties in the settings are expected to be specified as JSON objects. Please ensure the values are JSON objects.
```
jimmyca15, line 9:
```suggestion
  - The behavior when importing configurations with JSON content types such as `application/json` or `application/vnd.mycustomresource+json` has changed. The properties in the provided configuration are serialized as JSON to respect the provided content type. Settings that are expected to end up as JSON serialized strings in App Configuration should now be specified as JSON objects.

```

jimmyca15 (approved):

PR #773, opened 2023-06-07
2.6.0 preview2 release notes

rossgrambo:
Applications using built-in `ConfigurationFeatureDefinitionProvider` will now benefit from caching of feature filter parameters. This will improve performance of the application by reducing the number of times a filter's parameters are cast in short time frames, yielding observed performance increase of up to 100x. This change will not affect custom filters by default. For custom filters, the class must implement the `IFilterParametersBinder` interface. Below is an example.

```csharp
class MyFilter : IFeatureFilter, IFilterParametersBinder
{
    public object BindParameters(IConfiguration filterParameters)
    {
        return filterParameters.Get<FilterSettings>();
    }

    public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context, CancellationToken cancellationToken)
    {
        FilterSettings settings = (FilterSettings)context.Settings;
        ...
    }
}
```

For more details read [here](https://github.com/microsoft/FeatureManagement-Dotnet/pull/229)


releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -4,6 +4,30 @@
 # Microsoft.FeatureManagement.AspNetCore
 [Source code ][source_code_web] | [Package (NuGet)][package_web] | [Samples][samples_web] | [Product documentation][docs]
 
+## 2.6.0-preview2 - June 7, 2023
+
+### Enhancement - Parameter Caching
+
+Developers using our `ConfigurationFeatureDefinitionProvider` will now benefit from caching of filter parameters. This will improve performance of the library by reducing the number of times a filter's parameters are cast in short time frames, allowing loops to execute up to 100x faster. This change will not affect custom filters by default. For custom filters, the class must implement the `IFilterParametersBinder` interface. Below is an example.
```
jimmyca15, line 11: > allowing loops to execute up to 100x faster

It's somewhat odd to see loops being called out explicitly, given they are a bit of a contrived performance scenario. I'd suggest to have something like "yielding observed performance increase of up to 100x."

jimmyca15 (approved):

releaseNotes/Microsoft.Featuremanagement.md
zhenlan, line 11: > Developers using our ...

Applications using built-in ...

> improve performance of the library

improve performance of the application...

rossgrambo: I wanted to call out that the 100x performance increase is only true if calling IsEnabledAsync repeatedly. As calling it once every 1 second has a negligible performance increase. I suppose "short time frame" indicates that. I'll go with your suggestion.

PR #785, opened 2023-07-17
add a simple project that can be used in portal quick start blade

ivywei0125:
Add a simple console project. 
This project is used in portal quick start view. Uses could following steps in portal quick start view, download this project in step2 and connect to his own store with this project.
Snipaste_2023-07-25_17-23-21


examples/DotNetCore/SimpleConsoleApp/Program.cs
```diff
@@ -0,0 +1,18 @@
...
+
+    class Program
+    {
+        static void Main(string[] args) {
+
+            var builder = new ConfigurationBuilder();
+            builder.AddAzureAppConfiguration(Environment.GetEnvironmentVariable("ConnectionString"));
+
+            var config = builder.Build();
+            Console.WriteLine(config["TestApp:Settings:Message"] ?? "Hello world!");
```
zhenlan, line 13: Instead of "Hello world!", I would put something like below if the key is not found.

$"Please create a key-value with the key '{key}' in Azure App Configuration."

examples/DotNetCore/SimpleConsoleApp/SimpleConsoleApp.csproj
zhenlan, line 1: How about we call this "QuickStartApp"? We can then do something similar for other languages in the future.

examples/DotNetCore/SimpleConsoleApp/Program.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace Microsoft.Extensions.Configuration.AzureAppConfiguration.Examples.SimpleConsoleApp
+{
+    using Microsoft.Extensions.Configuration;
+
+    class Program
+    {
+        static void Main(string[] args) {
```
zhenlan, line 7: Can we create the project using the latest template that doesn't have the main function etc?

examples/DotNetCore/SimpleConsoleApp/Program.cs
ivywei0125: Sonds great! Updated.

examples/DotNetCore/SimpleConsoleApp/SimpleConsoleApp.csproj
ivywei0125: Make sense. Updated.

examples/DotNetCore/SimpleConsoleApp/Program.cs
ivywei0125: Updated.

avanigupta (approved):
Looks good! 
Is the key name in portal screenshot configurable by users? If not, we should change the key being selected in this sample (QuickStartApp*) to the one being created by default in the portal (TestApp*).

ivywei0125:
@avanigupta Yes, the key value should be created by users. We suggest users to create a key value {QuickStartApp:Settings:Message: Hello from App Configuration!} in portal. 
I have updated the screen shot to avoid misunderstanding.

PR #803, opened 2023-09-20
update link and keep old version schema

Eskibear:
Schema link in readme is broken, pointing to v1.0.0 which has been deleted. We probably should keep schema of old version for reference when we publish new version. 

This PR restore the old v1.0.0 schema file, and update the link pointing to the latest (v1.1.0) one.


zhenlan (approved):
Thanks for fixing this, @Eskibear

PR #806, opened 2023-09-29
Updating Python Samples to use Refresh

mrm9084:


examples/Python/python-django-webapp-sample/README.md
```diff
@@ -68,6 +69,19 @@ python manage.py migrate
...
+To refresh your configuration, you first update the value in Azure App Configuration, then update the Sentinel value to trigger a refresh.
+
+```Powershell
+az appconfig kv set --name <config-store-name> --key testapp_settings_color --value "violet"
+az appconfig kv set --name <config-store-name> --key sentinel --value "2"
+```
+
+Refresh the page in your browser to see the updated value.
+
+NOTE: By default refresh can only be triggered every 30 seconds. It may have to wait up till 30 seconds and then refreshing a page to see a refresh.
```
albertofori, line 83: Should this be `You may have to wait up to 30 seconds before refreshing the page to see the changes.`?

examples/Python/python-flask-webapp-sample/README.md
```diff
@@ -65,6 +66,19 @@ Start the flask application using the flask command:
...
+To refresh your configuration, you first update the value in Azure App Configuration, then update the Sentinel value to trigger a refresh.
+
+```Powershell
+az appconfig kv set --name <config-store-name> --key testapp_settings_color --value "violet"
+az appconfig kv set --name <config-store-name> --key sentinel --value "2"
+```
+
+Refresh the page in your browser to see the updated value.
+
+NOTE: By default refresh can only be triggered every 30 seconds. It may have to wait up till 30 seconds and then refreshing a page to see a refresh.
```
albertofori, line 80: Same here. Should this be updated accordingly?

examples/Python/python-django-webapp-sample/README.md
mrm9084: The thing I want to make clear here is that you have to wait 30 seconds, then refresh the page. Just waiting 30 will do nothing.

albertofori (approved):

PR #815, opened 2023-10-18
Include depth in example of breaking change importing configurations with JSON content types

MaryanneNjeri:
Include the depth property  in example of breaking change when importing configurations with JSON content types


releaseNotes/AzureDevOpsPushPipelineExtension.md
```diff
@@ -35,6 +35,7 @@ Azure App Configuration Push extension for Azure DevOps pipeline can be installe
     inputs:
       azureSubscription: '<subscription>'
       AppConfigurationEndpoint: '<store endpoint>'
+      Depth: 1
```
jimmyca15, line 38: This can easily be lost. Seems the sentence above that mentions to use a json object instead should also call out that depth needs to be updated.

MaryanneNjeri: How about something like "Additionally, the depth property should also be adjusted to match the expected depth of the JSON structure."

jimmyca15 (approved):

PR #841, opened 2023-12-01
Add Feature Flag Schema 2.0.0

amerjusupovic:


docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
...
+        "type": "boolean",
+        "title": "Enabled State",
+        "description": "A feature is OFF if enabled is false. If enabled is true, then the feature is ON if there are no conditions (null or empty) or if all conditions are satisfied.",
+        "default": false
+      },
+      "status": {
+        "$id": "#/properties/status",
+        "type": "string",
+        "title": "Feature Status",
+        "description": "Describes how the feature's enabled state is evaluated. When equal to Conditional, the state is conditional upon feature evaluation. When equal to Disabled, the feature will always be considered disabled.",
```
zhiyuanliang-ms, line 56:
```suggestion
        "description": "Describes how the feature's enabled state is evaluated. When set to 'Conditional', the state is conditional upon feature evaluation. When set to 'Disabled', the feature will always be considered disabled.",
```

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
...
+                    "title": "Variant Name",
+                    "description": "The name used to refer to a feature variant.",
+                    "default": "",
+                    "pattern": "^(.*)$"
+                },
+                "configuration_value": {
+                    "$id": "#/properties/variants/items/properties/configuration_value",
+                    "type": ["string", "null", "number", "object", "array", "boolean"],
+                    "title": "Variant Configuration Value",
+                    "description": "The inline value set for this feature variant."
```
zhiyuanliang-ms, line 168: Do we need to specify the pattern for `configuration_value` here (like `parameters` of `client_filters`)?

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
...
+                "title": "Variant Allocation Default When Enabled",
+                "description": "Specifies which variant should be used when a variant is requested while the feature is considered enabled and no other variant was allocated to the user.",
+                "default": "",
+                "pattern": "^(.*)$"
+            },
+            "user": {
+                "$id": "#/properties/allocation/properties/user",
+                "type": "array",
+                "title": "User Allocation Collection",
+                "description": "A list of objects each containing a variant name and list of users for which that variant should be used.",
```
zhiyuanliang-ms, line 220:
```suggestion
                "description": "A list of objects, each containing a variant name and list of users for whom that variant should be used.",
```

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
...
+                            "default": "",
+                            "pattern": "^(.*)$"
+                        },
+                        "users": {
+                            "$id": "#/properties/allocation/properties/user/items/properties/users",
+                            "type": "array",
+                            "title": "User Allocation Users Collection",
+                            "description": "Collection of users where if any match the current user, the variant specified in the user allocation is used.",
+                            "items": {
+                                "type": "string"
```
zhiyuanliang-ms, line 244: Do we need to specify the pattern here?

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
...
+                            }
+                        }
+                    }
+                }
+            },
+            "group": {
+                "$id": "#/properties/allocation/properties/group",
+                "type": "array",
+                "title": "Group Allocation Collection",
+                "description": "A list of objects each containing a variant name and list of groups for which that variant should be used.",
```
zhiyuanliang-ms, line 254:
```suggestion
                "description": "A list of objects, each containing a variant name and list of groups for which that variant should be used.",
```

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
...
+                            }
+                        }
+                    }
+                }
+            },
+            "percentile": {
+                "$id": "#/properties/allocation/properties/percentile",
+                "type": "array",
+                "title": "Percentile Allocation Collection",
+                "description": "A list of objects each containing a variant name and percentage range for which that variant should be used.",
```
zhiyuanliang-ms, line 288:
```suggestion
                "description": "A list of objects, each containing a variant name and percentage range for which that variant should be used.",
```

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
...
+                "default": "",
+                "pattern": "^(.*)$"
+            }
+        }
+      },    
+      "telemetry_enabled": {
+        "$id": "#/properties/telemetry_enabled",
+        "type": "boolean",
+        "title": "Feature Telemetry Enabled State",
+        "description": "A flag to enable or disable sending telemetry events.",
```
zhiyuanliang-ms, line 338: Should we drop the word "events"? I think "sending telemetry" is enough.

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
...
+        "default": false
+      },
+      "telemetry_metadata": {
+        "$id": "#/properties/telemetry_metadata",
+        "type": "object",
+        "title": "Feature Telemetry Metadata",
+        "description": "A container for metadata relevant to telemetry.",
+        "patternProperties": {
+            "^.*$": {
+              "type": "string"
```
zhiyuanliang-ms, line 348: Do we prohibit properties being number or boolean here?

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
amerjusupovic: From what I understand, `parameters` always expects an object so it makes sense to have the `patternProperties`, but I didn't know if I need it for this case if it's sometimes an object, but `configuration_value` can be anything.

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
amerjusupovic: In the feature definition `TelemetryMetadata` is a `Dictionary<string, string>`, so I thought it might be the same here.

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
...
+        }
+      },    
+      "telemetry_enabled": {
+        "$id": "#/properties/telemetry_enabled",
+        "type": "boolean",
+        "title": "Feature Telemetry Enabled State",
+        "description": "A flag to enable or disable sending telemetry events.",
+        "default": false
+      },
+      "telemetry_metadata": {
```
jimmyca15, line 341: This is not a desired property.

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
...
+        "pattern": "^(.*)$"
+      },
+      "enabled": {
+        "$id": "#/properties/enabled",
+        "type": "boolean",
+        "title": "Enabled State",
+        "description": "A feature is OFF if enabled is false. If enabled is true, then the feature is ON if there are no conditions (null or empty) or if all conditions are satisfied.",
+        "default": false
+      },
+      "status": {
```
zhenlan, line 52: This is a .NET only concept. It doesn't exist in AppConfig schema (which uses the "enabled" property).

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
...
+        "examples": [
+          "Fancy Background"
+        ],
+        "pattern": "^(.*)$"
+      },
+      "enabled": {
+        "$id": "#/properties/enabled",
+        "type": "boolean",
+        "title": "Enabled State",
+        "description": "A feature is OFF if enabled is false. If enabled is true, then the feature is ON if there are no conditions (null or empty) or if all conditions are satisfied.",
```
zhenlan, line 49: Since we added the requirement_type, can we make a small update to the description "...or if ~all~ conditions are satisfied". Can you make the same update to the 1.1.0 schema?

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
+{
+    "definitions": {},
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "http://azconfig.io/schemas/FeatureManagement/v2.0.0/FeatureFlag.json",
+    "type": "object",
+    "title": "An Azure App Configuration Feature Declaration",
+    "required": [
+      "id",
+      "enabled",
+      "conditions"
```
zhenlan, line 10: I understand this is from the old schema, but I don't think `conditions` is required. Can we remove it?

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
+{
+    "definitions": {},
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "http://azconfig.io/schemas/FeatureManagement/v2.0.0/FeatureFlag.json",
+    "type": "object",
+    "title": "An Azure App Configuration Feature Declaration",
+    "required": [
+      "id",
+      "enabled",
```
zhenlan, line 9: If the `enabled` property has a default value, it is not required. Do we require it in our provider/FM libraries?

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
...
+            "required": [
+                "name"
+            ],
+            "properties":{
+                "name": {
+                    "$id": "#/properties/variants/items/properties/name",
+                    "type": "string",
+                    "title": "Variant Name",
+                    "description": "The name used to refer to a feature variant.",
+                    "default": "",
```
zhenlan, line 161: It shouldn't have default if it's required.

amerjusupovic: Yea I see now that I mixed up the defaults on a few properties, I'll fix them.

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
...
+                    "type": ["string", "null", "number", "object", "array", "boolean"],
+                    "title": "Variant Configuration Value",
+                    "description": "The inline value set for this feature variant."
+                },
+                "configuration_reference": {
+                    "$id": "#/properties/variants/items/properties/configuration_reference",
+                    "type": "string",
+                    "title": "Variant Configuration Reference",
+                    "description": "The string used to reference a path in the configuration and set the variant's value equal to the value at that path.",
+                    "default": "",
```
zhenlan, line 175: Just to confirm, configuration_value doesn't have a default but configuration_reference does?

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
...
+                    "pattern": "^[a-zA-Z0-9]+(:[a-zA-Z0-9]+)*$"
+                },
+                "status_override": {
+                    "$id": "#/properties/variants/items/properties/status_override",
+                    "type": "string",
+                    "title": "Variant Status Override",
+                    "description": "Overrides the enabled state of the feature according to this value when this variant is used. Does not override the state if value is None.",
```
zhenlan, line 182: > ... if value is None.

Not sure what this means.

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
amerjusupovic: I think I swapped a lot of the defaults, if anything configuration_value default should be null and reference should have no default.

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
...
+        "type": "object",
+        "title": "Variant Allocation",
+        "description": "Determines how variants should be allocated for the feature to various users.",
+        "required": [],
+        "properties": {
+            "default_when_disabled": {
+                "$id": "#/properties/allocation/properties/default_when_disabled",
+                "type": "string",
+                "title": "Variant Allocation Default When Disabled",
+                "description": "Specifies which variant should be used when a variant is requested while the feature is considered disabled.",
```
zhenlan, line 204: Maybe we can simplify the sentence a little bit. Does this convey the same meaning?

Specifies which variant should be used when ~a variant is requested while~ the feature is considered disabled.

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
...
+                        "variant",
+                        "users"
+                    ],
+                    "properties": {
+                        "variant": {
+                            "$id": "#/properties/allocation/properties/user/items/properties/variant",
+                            "type": "string",
+                            "title": "User Allocation Variant",
+                            "description": "The name of the variant to use if the user allocation matches the current user.",
+                            "default": "",
```
zhenlan, line 235: No default

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
...
+                "title": "Variant Allocation Default When Disabled",
+                "description": "Specifies which variant should be used when a variant is requested while the feature is considered disabled.",
+                "default": "",
+                "pattern": "^(.*)$"
+            },
+            "default_when_enabled": {
+                "$id": "#/properties/allocation/properties/default_when_enabled",
+                "type": "string",
+                "title": "Variant Allocation Default When Enabled",
+                "description": "Specifies which variant should be used when a variant is requested while the feature is considered enabled and no other variant was allocated to the user.",
```
zhenlan, line 212: I'm trying to see if we can avoid the "user" concept here so people don't confuse it with the "user" section below. Does this work?

Specifies which variant should be used when ~a variant is requested while~ the feature is considered enabled and no other allocation rules apply.

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
...
+                        "variant",
+                        "from",
+                        "to"
+                    ],
+                    "properties": {
+                        "variant": {
+                            "$id": "#/properties/allocation/properties/percentile/items/properties/variant",
+                            "type": "string",
+                            "title": "Percentile Allocation Variant",
+                            "description": "The name of the variant to use if the calculated percentage for the current user falls in the provided range.",
```
zhenlan, line 303: ... calculated percentile ...

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
...
+                            "default": "",
```
zhenlan, line 304: No default

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
...
+                            "pattern": "^(.*)$"
+                        },
+                        "from": {
+                            "$id": "#/properties/allocation/properties/percentile/items/properties/from",
+                            "type": "number",
+                            "title": "Percentile Allocation From",
+                            "description": "The lower end of the percentage range for which this variant will be used.",
+                            "default": 0
```
zhenlan, line 312: No default

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
...
+                        },
+                        "to": {
+                            "$id": "#/properties/allocation/properties/percentile/items/properties/to",
+                            "type": "number",
+                            "title": "Percentile Allocation To",
+                            "description": "The upper end of the percentage range for which this variant will be used.",
+                            "default": 100
```
zhenlan, line 319: No default

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
```diff
@@ -0,0 +1,354 @@
...
+                        }
+                    }
+                }
+            },
+            "seed": {
+                "$id": "#/properties/allocation/properties/seed",
+                "type": "string",
+                "title": "Percentile Allocation Seed",
+                "description": "The value which percentile calculations are based on. Calculated percentages are consistent across features if the user and seed are consistent.",
```
zhenlan, line 328: I guess this is only used if this is not null/empty?

zhenlan, line 328: Calculated percentiles...

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
zhenlan: I expect it is. We will insert this dynamically, but if someone hardcodes this in a feature flag, it will be honored.

jimmyca15: Why would a customer not use tags on the key-value?

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
amerjusupovic: Sure, bit more concise

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
amerjusupovic: Portal doesn't let you omit enabled, but the provider will just set it to false by default

zhenlan: If the library doesn't enforce it, it's not required. Looks like our [new schema support PR](https://github.com/microsoft/FeatureManagement-Dotnet/pull/319/files#diff-55378af224e45c1157f406ef1f2714ddf93ee9ac4f312f5d6a4e6cf0fd58e72bR339) doesn't require it either. 

We should remove it from the required and all FM libraries should treat a feature flag as disabled if the property is not present.

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
jimmyca15: Discussed with Zhenlan. If we were to have feature flags as a resource at the same level of key-values we would most likely have telemetry configuration on top of existing App Configuration paradigms such as tags and etag. Thus, it is reasonable to include it here not only for feature management client operability but also because it makes sense to exist on the server.

Although since we have multiple properties revolving around telemetry we discussed to switch the schema. Here is an example following the updated schema

{
    "telemetry": {
        "enabled": true,
        "metadata": {
            "foo": "bar"
        }
    }
}

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
amerjusupovic: Ok, our original variants documentation mentions adding it to the schema so I went off of that, but I can remove it. Does that mean the .NET provider is just going to use the `enabled` property to send `status` to featuremanagement?

zhenlan: The "status" is a .NET FM only property designed to match the "enabled" property in the AppConfig schema. This is why it's included in the .NET FM document. In other words, the .NET FM document (implicitly) maintains another schema that is only used by the .NET FM library. But as you know, the .NET FM library is now adding support for the AppConfig schema too. The AppConfig schema is the one we plan to add support in FM libraries of all languages.

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
```diff
@@ -0,0 +1,345 @@
...
+                  "type": "string",
+                  "title": "Variant Configuration Reference",
+                  "description": "The string used to reference a path in the configuration and set the variant's value equal to the value at that path.",
+                  "pattern": "^[a-zA-Z0-9]+(:[a-zA-Z0-9]+)*$"
+              },
+              "status_override": {
+                  "$id": "#/properties/variants/items/properties/status_override",
+                  "type": "string",
+                  "title": "Variant Status Override",
+                  "description": "Overrides the enabled state of the feature according to this value when this variant is used. Does not override the state if value is None.",
```
zhenlan, line 168: How do you like this?

Overrides the enabled state of the feature if the given variant is assigned.

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
```diff
@@ -0,0 +1,345 @@
...
+                  "type": "string",
+                  "title": "Variant Name",
+                  "description": "The name used to refer to a feature variant.",
+                  "pattern": "^(.*)$"
+              },
+              "configuration_value": {
+                  "$id": "#/properties/variants/items/properties/configuration_value",
+                  "type": ["string", "null", "number", "object", "array", "boolean"],
+                  "title": "Variant Configuration Value",
+                  "description": "The inline value set for this feature variant.",
```
zhenlan, line 154: The configuration value for this feature variant.

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
```diff
@@ -0,0 +1,345 @@
...
+                  "default": null
+              },
+              "configuration_reference": {
+                  "$id": "#/properties/variants/items/properties/configuration_reference",
+                  "type": "string",
+                  "title": "Variant Configuration Reference",
+                  "description": "The string used to reference a path in the configuration and set the variant's value equal to the value at that path.",
```
zhenlan, line 161: The prefix of a configuration section used as the value for this feature variant.

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
```diff
@@ -0,0 +1,345 @@
...
+            }
+          }
+        }
+      }
+    },
+    "variants": {
+      "$id": "#/properties/variants",
+      "type": "array",
+      "title": "Variants Collection",
+      "description": "The list of variants defined for this feature. A variant represents a value of a feature flag that can be a string, a number, a boolean, or even a configuration object.",
```
zhenlan, line 134: A variant represents a **configuration** value of a feature flag that can be a string, a number, a boolean, or ~even~ a configuration object.

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
```diff
@@ -0,0 +1,345 @@
...
+      "$id": "#/properties/allocation",
+      "type": "object",
+      "title": "Variant Allocation",
+      "description": "Determines how variants should be allocated for the feature to various users.",
+      "required": [],
+      "properties": {
+          "default_when_disabled": {
+              "$id": "#/properties/allocation/properties/default_when_disabled",
+              "type": "string",
+              "title": "Variant Allocation Default When Disabled",
```
zhenlan, line 189: How about this?

Default Variant Allocation When Disabled

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
```diff
@@ -0,0 +1,345 @@
...
+              "description": "Specifies which variant should be used when the feature is considered disabled.",
+              "default": "",
+              "pattern": "^(.*)$"
+          },
+          "default_when_enabled": {
+              "$id": "#/properties/allocation/properties/default_when_enabled",
+              "type": "string",
+              "title": "Variant Allocation Default When Enabled",
```
zhenlan, line 197: Default Variant Allocation When Enabled

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
```diff
@@ -0,0 +1,345 @@
...
+                      "variant": {
+                          "$id": "#/properties/allocation/properties/percentile/items/properties/variant",
+                          "type": "string",
+                          "title": "Percentile Allocation Variant",
+                          "description": "The name of the variant to use if the calculated percentile for the current user falls in the provided range.",
+                          "pattern": "^(.*)$"
+                      },
+                      "from": {
+                          "$id": "#/properties/allocation/properties/percentile/items/properties/from",
+                          "type": "number",
```
zhenlan, line 292: QQ: is there a way to specify the range?

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
```diff
@@ -0,0 +1,345 @@
...
+                          "description": "The upper end of the percentage range for which this variant will be used."
+                      }
+                  }
+              }
+          },
+          "seed": {
+              "$id": "#/properties/allocation/properties/seed",
+              "type": "string",
+              "title": "Percentile Allocation Seed",
+              "description": "The value which percentile calculations are based on. Calculated percentiles are consistent across features if the user and seed are consistent.",
```
zhenlan, line 309: "The value, which...." or "The value percentile..."

"Calculated percentile is consistent across features for a given user if the same nonempty seed is used."

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
```diff
@@ -0,0 +1,345 @@
...
+              "default": "",
+              "pattern": "^(.*)$"
+          }
+      }
+    },
+    "telemetry": {
+      "$id": "#/properties/telemetry",
+      "type": "object",
+      "title": "Feature Telemetry Options",
```
zhenlan, line 318: We don't always prefix "Feature" for other top level properties. Feels redundant here.

~Feature~ Telemetry Options

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
```diff
@@ -0,0 +1,345 @@
...
+      "description": "The declaration of options used to configure telemetry for this feature.",
+      "required": [],
+      "properties": {
+        "enabled": {
+          "$id": "#/properties/telemetry/properties/enabled",
+          "type": "boolean",
+          "title": "Telemetry Enabled State",
+          "description": "A flag to enable or disable sending telemetry.",
```
zhenlan, line 326: The flag itself doesn't enable/disable telemetry. The flag is for telemetry publisher to consume, if installed. How about this?

Indicates if the telemetry of the feature is enabled.

docs/FeatureManagement/FeatureFlagv2.0.0.schema.json
amerjusupovic: Right, otherwise it uses a default value based on the feature name

zhenlan: Can we add clarification?

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
```diff
@@ -0,0 +1,345 @@
...
+          "default": false
+        },
+        "metadata": {
+          "$id": "#/properties/telemetry/properties/metadata",
+          "type": "object",
+          "title": "Telemetry Metadata",
+          "description": "A container for metadata relevant to telemetry.",
```
jimmyca15, line 333:
```suggestion
          "description": "A container for metadata that should be bundled with flag telemetry.",
```

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
```diff
@@ -0,0 +1,345 @@
...
+                  ],
+                  "properties": {
+                      "variant": {
+                          "$id": "#/properties/allocation/properties/percentile/items/properties/variant",
+                          "type": "string",
+                          "title": "Percentile Allocation Variant",
+                          "description": "The name of the variant to use if the calculated percentile for the current user falls in the provided range.",
+                          "pattern": "^(.*)$"
+                      },
+                      "from": {
```
jimmyca15, line 290: Should have min: 0, max: 100. Same with to.

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
jimmyca15: I would simplify further to "Indicates if telemetry is enabled."

jimmyca15 (approved):

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
```diff
@@ -0,0 +1,345 @@
...
+        },
+        "client_filters": {
+          "$id": "#/properties/conditions/properties/client_filters",
+          "type": "array",
+          "title": "Client Filter Collection",
+          "description": "Filters that must run on the client and be evaluated as true for the feature to be considered enabled.",
+          "items": {
+            "$id": "#/properties/conditions/properties/client_filters/items",
+            "type": "object",
+            "title": "Client Filter",
```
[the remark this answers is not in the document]
amerjusupovic: I saw this in the v1.1.0 file and I interpreted it as a singular filter because it's describing how each individual filter in the list should look, and there's already a `Client Filter Collection` object.

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
amerjusupovic: Right, since we agreed to use the term "assigned" now I think this makes it clearer.

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
amerjusupovic: Not sure what you mean, like instead of using `from` and `to` as the range?

zhenlan: The from and to must be in the range of 0 and 100. Is there a way to specify it?

amerjusupovic: Oh right, I think jimmy commented mentioning I should do that with `minimum` and `maximum` for those values. I've added that in the last commit.

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
```diff
@@ -130,8 +129,8 @@
     "variants": {
       "$id": "#/properties/variants",
       "type": "array",
-      "title": "Variants Collection",
-      "description": "The list of variants defined for this feature. A variant represents a value of a feature flag that can be a string, a number, a boolean, or even a configuration object.",
+      "title": "Variant Collection",
+      "description": "The list of variants defined for this feature. A variant represents a configuration value of a feature flag that can be a string, a number, a boolean, or even a configuration object.",
```
zhenlan, line 133: NIT: no need to stress on configuration object "... or ~even~ a configuration object".

zhenlan (approved):

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
zhenlan: A second thought, should it be "JSON object"? We don't support the configuration object to be anything else, do we?

amerjusupovic: In terms of how it's defined in app config/local files, it should always be a JSON. I can change it to say that

zhenlan (approved):

PR #844, opened 2023-12-06
Python with callback Samples

mrm9084:
Updates samples to use refresh and callbacks.

### DIFF
--- a/examples/Python/python-django-webapp-sample/quickstartproject/settings.py
+++ b/examples/Python/python-django-webapp-sample/quickstartproject/settings.py
@@ -13,33 +13,34 @@ https://docs.djangoproject.com/en/4.0/ref/settings/
 import os
 from pathlib import Path
 import configparser
-from azure.appconfiguration.provider import load, AzureAppConfigurationKeyVaultOptions, SettingSelector
+from azure.appconfiguration.provider import load, SettingSelector, SentinelKey
 from azure.identity import DefaultAzureCredential
 
 c_parser = configparser.ConfigParser()
 c_parser.read('static/config.ini')
 
-config = c_parser['DEFAULT']
+CONFIG = c_parser['DEFAULT']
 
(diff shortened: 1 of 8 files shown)


examples/Python/python-django-webapp-sample/quickstartproject/settings.py
```diff
@@ -30,17 +30,25 @@
...
-              keyvault_credential=credential,
-              credential=credential,
-              selects=[selects, selects_secret],
-              trim_prefixes=["testapp_settings_"],
-              refresh_on=[SentinelKey("sentinel")]
-              )
+
+def callback():
+    # Update Django settings with the app configuration key-values
+    CONFIG.update(AZURE_APP_CONFIG)
```
albertofori, line 36: Should this be inline on line 44? 

`on_refresh_success=lambda: CONFIG.update(AZURE_APP_CONFIG)`

examples/Python/python-flask-webapp-sample/app.py
```diff
@@ -1,46 +1,53 @@
...
+azure_app_config = None  # declare azure_app_config as a global variable
+
+def callback():
+   app.config.update(azure_app_config)
 
-if __name__ == '__main__':
-   app.run()
+async def load_config():
+   global azure_app_config
+   async with await load(endpoint=ENDPOINT,
```
albertofori, line 22: Is there a reason we use a `with` statement here?

examples/Python/python-django-webapp-sample/quickstartproject/settings.py
mrm9084: @albertofori, I was thinking about not doing that as this keeps it simple. And shows you can do what ever you want. Plus the comment is easy to map to the callback.

rossgrambo (approved):

PR #859, opened 2024-01-11
FeatureManagement v4 preview - release notes

rossgrambo:


releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -4,6 +4,22 @@
...
+## 4.0.0-preview - January 4, 2024
+
+### Breaking Changes
+
+There are no breaking changes in this release.
+
+### Enhancements
+
+* [Variants](https://github.com/microsoft/FeatureManagement-Dotnet/tree/release/v4?tab=readme-ov-file#variants) - Allows for more complex control of features than a simple on/off flag.
+* [Telemetry](https://github.com/microsoft/FeatureManagement-Dotnet/tree/release/v4?tab=readme-ov-file#telemetry) - Allows for analysis and tracking of feature assignment.
```
zhiyuanliang-ms, line 16:
```suggestion
* [Telemetry](https://github.com/microsoft/FeatureManagement-Dotnet/tree/release/v4?tab=readme-ov-file#telemetry) - Allows for analysis and tracking of feature flag state and variants assignment. 
```

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -4,6 +4,22 @@
...
+
+### Enhancements
+
+* [Variants](https://github.com/microsoft/FeatureManagement-Dotnet/tree/release/v4?tab=readme-ov-file#variants) - Allows for more complex control of features than a simple on/off flag.
+* [Telemetry](https://github.com/microsoft/FeatureManagement-Dotnet/tree/release/v4?tab=readme-ov-file#telemetry) - Allows for analysis and tracking of feature assignment. 
+* [Status](https://github.com/microsoft/FeatureManagement-Dotnet/tree/release/v4?tab=readme-ov-file#status) - A new optional field on a Feature that controls how a flag's enabled state is evaluated. Flags can set this field to "Disabled" which will cause the flag to always act disabled, regardless of filters or other schema. 
+
+### Additional
+
+* An example was added to demonstrate how to use the new Telemetry in an ASP.NET application. See https://github.com/microsoft/FeatureManagement-Dotnet/tree/release/v4?tab=readme-ov-file#telemetry
```
zhiyuanliang-ms, line 21:
```suggestion
* An example was added to demonstrate how to use the new Telemetry function in an ASP.NET application. See https://github.com/microsoft/FeatureManagement-Dotnet/tree/release/v4?tab=readme-ov-file#telemetry
```

zhiyuanliang-ms: The link seems to be duplicated here. Can we provide a link to the example or just remove this link?

rossgrambo: Oh oops, that's the wrong link. Thank you.

rossgrambo:
This PR has been significantly rewritten.

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -4,6 +4,99 @@
...
+
+```json
+"MyFlag": {
+    "Telemetry": {
+        "Enabled": true
+    }
+}
+```
+
+And a telemetry publisher needs to be registered. One is offered out-of-the-box for Application Insights and could be added with a single line.
```
jimmyca15, line 70: Rather than say out-of-the-box can you mention that it's available in the new app insights package.

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -4,6 +4,99 @@
 # Microsoft.FeatureManagement.AspNetCore
 [Source code ][source_code_web] | [Package (NuGet)][package_web] | [Samples][samples_web] | [Product documentation][docs]
 
+## 4.0.0-preview - January 4, 2024
+
+### Variants
+
+Variants provide the ability to surface different variations of a feature, allowing different segments of an audience to get different variations of a feature, all within a single feature flag. Variants also contain a value that can either reference another ConfigurationSection or be declared inline on the flag.
```
jimmyca15, line 11: Repition of the phrase "variations of a feature" breaks the flow of the sentence for me.

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -4,6 +4,99 @@
...
+    "Percentile": []
+}
+```
+
+A feature with Variants will still have a boolean result when called via `IsEnabledAsync`, following the usual filters. However, when a feature is called with `GetVariantAsync`, it will evaluate the allocation and return a `Variant` object. 
+
+```csharp
+Variant variant = await featureManager.GetVariantAsync(MyFeatureFlags.FeatureU);
+
+String variation = variant.Name;
```
jimmyca15, line 50: in c# string should be used as `string`

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -4,6 +4,99 @@
...
+},  
+{ 
+    "Name": "Small", 
+    "ConfigurationValue": {
+        "Size": 300
+    }
+}
+```
+
+Variants are then assigned via Allocation, which is defined as a part of the feature flag. The fields are User, Group, Percentile, and a couple defaults.
```
jimmyca15, line 26: It's a delicate balance between to create a sneak peek. I think that going over the allocation schema might be a bit much and should be left to the next step of a readme visit.

releaseNotes/Microsoft.Featuremanagement.md
jimmyca15: I took a look at the dynamic features opening section, it looks like it could have been used here without much modification. It also mentions one of the key points that a variant is not limited to boolean.

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -4,6 +4,99 @@
...
+
+### Additional Changes
+
+#### IVariantFeatureManager
+
+`IVariantFeatureManager` has been added as the successor of the existing `IFeatureManager`. It continues to offer the functions of `IFeatureManager`, but offers the new `GetVariantAsync` methods as well.
+
+#### Cancellation Tokens
+
+`IVariantFeatureManager` incorporates cancellation tokens into the methods of `IFeatureManager`. To migrate from the old interface, simply use the new one, and adjust calls to `IsEnabledAsync` or `GetFeatureNamesAsync` to include a `CancellationToken`.
```
jimmyca15, line 90: I wouldn't dub either interface old or new, but rather say to take advantage, use `IVariantFeatureManager` instead of IFeatureManager. If we were to own up to 'old' moniker we should deprecate.

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -4,6 +4,99 @@
...
+
+String variation = variant.Name;
+IConfigurationSection variantConfiguration = variant.Configuration;
+```
+
+For more details on Variants, see [here](https://github.com/microsoft/FeatureManagement-Dotnet/tree/release/v4?tab=readme-ov-file#variants).
+
+### Telemetry
+
+Telemetry offers the ability to emit events containing information about a feature evaluation. This can be used to ensure a flag is running as expected, or to see which users were given which features and why they were given the feature. To enable this functionality, two things need to be done.
```
jimmyca15, line 58:
```suggestion
The feature management library now offers the ability to emit events containing information about a feature evaluation. This can be used to ensure a flag is running as expected, or to see which users were given which features and why they were given the feature. To enable this functionality, two things need to be done.
```

releaseNotes/Microsoft.Featuremanagement.md
rossgrambo: Reworked it around your other comments as well. Still included a small snippet about allocation, let me know what you think.

releaseNotes/Microsoft.Featuremanagement.md
rossgrambo: Done. I adjusted it a little extra around the "Previously, this library only worked with feature flags." to try to separate "Feature Flags" from the idea of "flagging" or "boolean flags", since this library still "only works with Feature Flags", the proper-noun version.

Let me know if the adjusted version makes sense to you.

rossgrambo:
I switched the first json section to be code highlighted as javascript, which allows us to place comments in the json. Let me know if y'all think that's beneficial or you prefer valid JSON. Just hard to strike a balance with valid JSON between adding too much schema or omitting important context.

jimmyca15 (approved):

PR #871, opened 2024-02-06
Add Feature Management schema

zhiyuanliang-ms:
This feature management schema will be provided by the Kubernetes provider and it would be recognized by client Feature Management libraries. It would be better to place the schema in the App Config repo instead of the Feature Management repo, as discussed [here](https://github.com/microsoft/FeatureManagement-Dotnet/pull/362#discussion_r1479058573)


docs/FeatureManagement/FeatureManagement.v1.0.0.schema.json
```diff
@@ -0,0 +1,31 @@
+{
+    "definitions": {},
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "http://azconfig.io/schemas/FeatureManagement/v1.0.0/FeatureManagement.json",
+    "type": "object",
+    "title": "An Azure App Configuration Feature Management Configuration",
```
jimmyca15, line 6: I would say this is Microsoft Feature Management schema.

The schema that describes individual flag is the `Microsoft Feature Flag Schema`.

docs/FeatureManagement/FeatureManagement.v1.0.0.schema.json
```diff
@@ -0,0 +1,31 @@
...
+                "FeatureFlags"
+            ],
+            "properties": {
+                "FeatureFlags": {
+                    "$id": "#/properties/FeatureManagement/properties/FeatureFlags",
+                    "type": "array",
+                    "title": "Feature Flags",
+                    "description": "Declares feature flags based on Microsoft Feature Flag schema.",
+                    "items": {
+                        "$ref": "http://azconfig.io/schemas/FeatureManagement/v1.1.0/FeatureFlag.json"
```
rossgrambo, line 25: $ref should be the uri of the file rather than the $id. I believe `./FeatureFlag.v1.1.0.schema.json` would work, but to make it work without folder context- I think we should do: `https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json`

docs/FeatureManagement/FeatureManagement.v1.0.0.schema.json
zhiyuanliang-ms: I like the idea naming this schema as `Microsoft Feature Management`.
@zhenlan 

As discussed in the email thread, snake_casing will be used:
```json
"feature_management": {
  "feature_flags": [
     {
       // Microsoft feature flag schema
     }
  ]
}

```

docs/FeatureManagement/FeatureManagement.v1.0.0.schema.json
```diff
@@ -0,0 +1,31 @@
...
+    "type": "object",
+    "title": "A Microsoft Feature Management Configuration",
+    "required": [
+        "feature_management"
+    ],
+    "properties": {
+        "feature_management":{
+            "$id": "#/properties/feature_management",
+            "type": "object",
+            "title": "The Feature Management Schema",
```
RichardChen820, line 14: Need a description for it?

RichardChen820 (approved):

docs/FeatureManagement/FeatureManagement.v1.0.0.schema.json
zhiyuanliang-ms: added

Eskibear (approved):

docs/FeatureManagement/FeatureManagement.v1.0.0.schema.json
```diff
@@ -0,0 +1,32 @@
+{
+    "definitions": {},
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "http://azconfig.io/schemas/FeatureManagement/v1.0.0/FeatureManagement.json",
+    "type": "object",
+    "title": "A Microsoft Feature Management Configuration",
```
zhiyuanliang-ms, line 6:
```suggestion
    "title": "The Microsoft Feature Management Schema",
```

jimmyca15 (approved):

PR #885, opened 2024-03-11
Update description of "requirement_type“ for feature flag schemas

zhiyuanliang-ms:
"The filter is enabled" sounds like the filter is working/being evaluated. The emphasis here is the result of filter evaluation.

https://github.com/microsoft/FeatureManagement-JavaScript/pull/1#discussion_r1517061416


Eskibear (approved):
LGTM

PR #886, opened 2024-03-14
Release notes for 4.0.0-preview2

rossgrambo:


releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -4,6 +4,90 @@
 # Microsoft.FeatureManagement.AspNetCore
 [Source code ][source_code_web] | [Package (NuGet)][package_web] | [Samples][samples_web] | [Product documentation][docs]
 
+## 4.0.0-preview2 - March 7, 2024
+
+### Feature Service Injector
```
jimmyca15, line 9:
```suggestion
### Use flags to enable different service implementations in dependency injection
```

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -4,6 +4,90 @@
...
+
+Dependency injection can be wired up with a variant feature flag. Different implementations of a service interface can be considered as variant services. Using the newly provided `IVariantServiceProvider<TService>`, the appropriate service can be injected.
```
jimmyca15, line 11: I'd suggest opening with something like "It is now possible to use variant feature flags to control which implementation of a service dependency injection uses."

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -4,6 +4,90 @@
...
+
+Lastly, the provider is ready to be used and grab the appropriate service for the variant.
+
+``` C#
+IVariantServiceProvider<IAlgorithm> algorithmServiceProvider;
+...
+IAlgorithm forecastAlgorithm = await algorithmServiceProvider.GetServiceAsync(cancellationToken);
+```
+
+The readme will be updated to fully describe the new functionality.
```
jimmyca15, line 57: Good point. I think we need to get the readme in first and just point to that.

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -4,6 +4,90 @@
...
+
+### Telemetry & Targeting
+
+This preview release includes some additional mechanisms to track targeting within telemetry. There's been some changes to the telemetry publisher fields, and a couple classes are now offered/have been updated: `TargetingTelemetryInitializer` in `Microsoft.FeatureManagement.Telemetry.AspNetCore` and `TargetingHttpContextMiddleware` in `Microsoft.FeatureManagement.Telemetry.AspNetCore`.
```
jimmyca15, line 61: Seems `TargetingTelemetryInitializer` and `TargetingHttpContextMiddleware` need to be mentioned in our readme.

releaseNotes/Microsoft.Featuremanagement.md
rossgrambo: Agreed. Stole a lot from https://github.com/microsoft/FeatureManagement-Dotnet/pull/388/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R977 , but this should be more about announcing it rather than using it.

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -4,6 +4,90 @@
...
+IAlgorithm forecastAlgorithm = await algorithmServiceProvider.GetServiceAsync(cancellationToken);
+```
+
+The readme will be updated to fully describe the new functionality.
+
+### Telemetry & Targeting
+
+This preview release includes some additional mechanisms to track targeting within telemetry. There's been some changes to the telemetry publisher fields, and a couple classes are now offered/have been updated: `TargetingTelemetryInitializer` in `Microsoft.FeatureManagement.Telemetry.AspNetCore` and `TargetingHttpContextMiddleware` in `Microsoft.FeatureManagement.Telemetry.AspNetCore`. 
+
+These changes introduce the field `TargetingId`, which tracks the id used by TargetingFilters and Allocation. This allows telemetry to connect events by the target to the evaluation events for the target, rather than relying on built in app insights fields that may or may not exist in a project.
```
jimmyca15, line 63: > app insights

This change is unrelated to app insights, no?

releaseNotes/Microsoft.Featuremanagement.md
jimmyca15: Which has been added, which has been updated?

Suggest doing like

* Class (added)
* Class2 (updated)

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -4,6 +4,90 @@
...
+``` C#
+builder.Service.AddSingleton<ITelemetryInitializer, TargetingTelemetryInitializer>();
+
+app.UseMiddleware<TargetingHttpContextMiddleware>();
+```
+
+### Additional Changes
+
+#### BlazorServerApp
+
```
jimmyca15, line 76: That's not part of the released packages.

jimmyca15: These release notes are limited exclusively to the functionality of the released packages.

releaseNotes/Microsoft.Featuremanagement.md
zhenlan: Agreed. I would avoid turning our release notes into feature documentation. Mostly of these should go to proper documentation.

releaseNotes/Microsoft.Featuremanagement.md
zhenlan: We are explaining the design choice here? This is way too much for release notes. It should go to documentation.

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -4,6 +4,90 @@
...
+An example for Blazor has been added named `BlazorServerApp`. This app demonstrates how to handle FeatureMangement setup, use, and targeting context accessors in Blazor.
+
+#### Schema
+
+A schema file has been introduced to explicitly declare [the Feature Flag schema](https://github.com/microsoft/FeatureManagement-Dotnet/blob/release/v4/schemas/FeatureManagement.Dotnet.v1.0.0.schema.json).
+
+#### Net8 build target
+
+Net8 was added as a build target.
```
zhenlan, line 85: This should be presented as a feature - we added support for .NET 8. 

(Reminder of the release notes guidelines - talk about the feature in customer perspective, not how we implement it.)

releaseNotes/Microsoft.Featuremanagement.md
zhiyuanliang-ms: > this should be more about announcing it rather than using it.

Agree. I will update the README.

rossgrambo:
This is blocked by https://github.com/microsoft/FeatureManagement-Dotnet/pull/388 for now.

releaseNotes/Microsoft.Featuremanagement.md
rossgrambo: Went with:

> It is now possible to use variant feature flags to control which implementation of a service is used in dependency injection. This enables different implementations to be used for different audiences of the application. After using the newly provided `WithVariantService<TService>` builder method, a `IVariantServiceProvider<TService>` can be injected which will supply the appropriate variation of `TService` determined by allocation.

And ripped out all code snippets about it.

releaseNotes/Microsoft.Featuremanagement.md
rossgrambo: Removed the paragraph

releaseNotes/Microsoft.Featuremanagement.md
rossgrambo: Hard to find the correct thing to say- seeing as the only change was adding it was a build target- which means we're less likely to break Net8, but there's no immediate customer benefit. 

I went with:

> ### Net8 support
> 
> Added support for Net8 applications by adding Net8 as a build target.

zhiyuanliang-ms: https://github.com/microsoft/FeatureManagement-Dotnet/pull/365
What this PR did was to have our test suites also build for .NET 8.0.

Maybe we should not mention this in the release note, since the user will not notice this change.

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -4,6 +4,37 @@
 # Microsoft.FeatureManagement.AspNetCore
 [Source code ][source_code_web] | [Package (NuGet)][package_web] | [Samples][samples_web] | [Product documentation][docs]
 
+## 4.0.0-preview2 - March 7, 2024
```
zhenlan, line 7: Looks like we are listing out every feature/change separately. Is there a reason we don't follow the release notes guidelines?

releaseNotes/Microsoft.Featuremanagement.md
rossgrambo: Now links to readme

releaseNotes/Microsoft.Featuremanagement.md
rossgrambo: Which guideline do you feel isn't followed? This is primarily built off previous releases in this library. 

Releases for FM have been more than just release notes, but have also acted as "marketing" for the new features. We usually list the major features, then list enhancements/additional changes, then breaking changes. 

In this PR specifically, comments caused me to remove all but one of the "enchancements", so I removed the header. https://github.com/Azure/AppConfiguration/pull/886/commits/e01c3f4bb2389b600e9dc7ef380cb9fc3ff2bffd#diff-062e587e3d06ace0b7748874ffc0d6022532d33be3d8199526fa7aaadd07030dR82

zhenlan: Shared the release notes guideline offline. Not a problem at all if we want to change it, but we should discuss it.

I saw the preview1 release notes didn't follow the guideline either. That's exactly the problem. Once we had something like that, people will follow it as precedence. This is exactly the problem the guideline tries to solve.

rossgrambo:
This has been updated to follow release notes guidelines.

jimmyca15 (approved):

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -4,6 +4,18 @@
...
+
+### Enchancements
+
+* Use flags to enable different service implementations in dependency injection. ([#39](https://github.com/microsoft/FeatureManagement-Dotnet/issues/39)). See more details [here](https://github.com/microsoft/FeatureManagement-Dotnet/tree/preview?tab=readme-ov-file#variants-in-dependency-injection).
+* Additional mechanisms to track targeting within telemetry to improve the connection between published events and metrics. ([#409](https://github.com/microsoft/FeatureManagement-Dotnet/issues/409))
+* Added support for Net8 applications by adding Net8 as a build target. ([#364](https://github.com/microsoft/FeatureManagement-Dotnet/issues/364))
```
zhenlan, line 13: It's not that we didn't support .NET 8 applications before this change. By adding .NET 8 to the targeting framework, we make sure this library uses .NET 8 libraries/APIs when users build a .NET 8 application.

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -4,6 +4,18 @@
 # Microsoft.FeatureManagement.AspNetCore
 [Source code ][source_code_web] | [Package (NuGet)][package_web] | [Samples][samples_web] | [Product documentation][docs]
 
+## 4.0.0-preview2 - March 7, 2024
+
+### Enchancements
```
jimmyca15, line 9: Enhancements

releaseNotes/Microsoft.Featuremanagement.md
rossgrambo: Based it off your previous comment: https://github.com/Azure/AppConfiguration/pull/886#discussion_r1525558270

We added net8 as a build target which means we will run tests against it. We did also update the external libraries from 7.0 to 8.0. The only customer facing change here is: "we're less likely to break a .net8 app now". I'd argue that's "adding support ...", but as you said that makes it sound like it wasn't supported at all before. What about just "Added better support ..."?

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -4,6 +4,18 @@
...
+
+* Use flags to enable different service implementations in dependency injection. ([#39](https://github.com/microsoft/FeatureManagement-Dotnet/issues/39)). See more details [here](https://github.com/microsoft/FeatureManagement-Dotnet/tree/preview?tab=readme-ov-file#variants-in-dependency-injection).
+* Additional mechanisms to track targeting within telemetry to improve the connection between published events and metrics. ([#409](https://github.com/microsoft/FeatureManagement-Dotnet/issues/409))
```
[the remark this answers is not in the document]
rossgrambo: This concept is a little in the weeds to dig into. The linked issue links to a PR that explains it https://github.com/microsoft/FeatureManagement-Dotnet/pull/350#issue-2077862243.

The customer does not receive much benefit from this- it was a change on how we connect dots which were already connectable. But this is a more stable version.

Do you prefer something like:

Introduced middleware and a telemetry property for Targeting Id, which can be used to connect published evaluation events to other emitted telemetry from a given target.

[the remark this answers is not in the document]
zhenlan: We released two new packages and found "The customer does not receive much benefit from this". We must be missing something :)

(We should create separate file(s) for the release notes.)

releaseNotes/Microsoft.Featuremanagement.md
rossgrambo: That works for me- maybe a small tweak to be:

Added support for .NET 8 target framework to the Microsoft.FeatureManagement.AspNetCore package.

PR #889, opened 2024-03-19
Add release notes for .NET provider release 8.0.0-preview.2

amerjusupovic:


releaseNotes/MicrosoftExtensionsConfigurationAzureAppConfiguration.md
```diff
@@ -1,6 +1,10 @@
 # Microsoft.Extensions.Configuration.AzureAppConfiguration
 [Source code][source_code] | [Package (NuGet)][package]
 
+## 8.0.0-preview.2 - March 18, 2024
+### Bug Fixes
+* Fixed a bug where the provider would throw a `FormatException` while parsing a feature flag if a variant's configuration value was not set to a string or `null`. [#531](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/531)
```
zhenlan, line 6: I normally try to avoid using "provider" without context. While it's clear to us, it's not always clear to customers what we refer to. Suggest below:

Fixed a bug where a `FormatException` may be thrown for a variant feature flag whose configuration value was not set to a string or `null`. With the fix, the configuration value of a variant can be a string, number, boolean, or JSON object.

zhenlan (approved):

PR #890, opened 2024-03-20
Feature Flag Schema Samples/Tests

mrm9084:
A json file with various versions of feature flags to validate our Feature Management libraries. The markdown file gives the expected results and a test description of what is going on.

Validated the results against Python and DotNet. Python has a PR up to fix bugs found by this.

Spring currently produces different targeting results; we will have to discuss how to make the update without it being a breaking change.


docs/FeatureManagement/FeatureFlag.v2.0.0.schema.sample.scenarios.md
```diff
@@ -0,0 +1,41 @@
...
+| 25     | Theta             | user="Brittney" groups=["Stage4"] | False           | Targeting Filter, group is not part of default rollout calculation, no change            |
+| 26     | Iota              | N/A                               | True            | Feature Flag with two feature filters, first returns true, so it's enabled.            |
+| 27     | Kappa             | N/A                               | True            | Feature Flag with two feature filters, second returns true, so it's enabled.            |
+| 28     | Lambda            | N/A                               | True            | Feature Flag with two feature filters and requirement type specified as Any. Second filter returns true            |
+| 29     | Mu                | N/A                               | False           | Feature Flag with two feature filters and requirement type specified as Any. Neither filter returns true            |
+| 30     | Nu                | user="Adam"                       | False           | Feature Flag with two feature filters and requirement type specified as All. Only the first fitler returns true          |
```
amerjusupovic, line 36:
```suggestion
| 30     | Nu                | user="Adam"                       | False           | Feature Flag with two feature filters and requirement type specified as All. Only the first filter returns true          |
```

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.sample.scenarios.md
```diff
@@ -0,0 +1,41 @@
...
+| 11     | Eta               | groups=["Stage2"]                 | True            | Targeting Filter, no user Stage 2 is 50% rolloued out.            |
+| 12     | Eta               | user="Adam" groups=["Stage2"]     | True            | Targeting Filter, Adam who is not part of the default rollout is part of the first 50% of Stage 2             |
+| 13     | Eta               | user="Chad" groups=["Stage2"]     | False           | Targeting Filter, Chad is neither part of the default rollout or part of the first 50% of Stage 2            |
+| 14     | Eta               | groups=["Stage3"]                 | False           | Targeting Filter, the Stage 3 is group is on the exclusion list            |
+| 15     | Eta               | user="Alice" groups=["Stage3"]    | False           | Targeting Filter, Alice who is targeted is excluded because she is part of the Stage 3 group           |
+| 16     | Eta               | user="Ellie" groups=["Stage3"]    | False           | Targeting Tilter Ellie who was enabled by the default rollout is now excluded as part of the Stage 3 group            |
```
amerjusupovic, line 22:
```suggestion
| 16     | Eta               | user="Ellie" groups=["Stage3"]    | False           | Targeting Filter, Ellie who was enabled by the default rollout is now excluded as part of the Stage 3 group            |
```

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.sample.scenarios.md
```diff
@@ -0,0 +1,41 @@
...
+| 28     | Lambda            | N/A                               | True            | Feature Flag with two feature filters and requirement type specified as Any. Second filter returns true            |
+| 29     | Mu                | N/A                               | False           | Feature Flag with two feature filters and requirement type specified as Any. Neither filter returns true            |
+| 30     | Nu                | user="Adam"                       | False           | Feature Flag with two feature filters and requirement type specified as All. Only the first fitler returns true          |
+| 31     | Xi                | user="Adam"                       | True            | Feature Flag with two feature filters and requirement type specified as All. Both filters return true            |
+
+## Extra Test Scenario
+
+Update the json to have the default rollout percentage of Theta to be 75% instead of 74% and run the test scenarios again. The expected result for tests 22-25 should now be True, this validates we are using the same fomula for our rollout calculation.
```
amerjusupovic, line 41:
```suggestion
Update the json to have the default rollout percentage of Theta to be 75% instead of 74% and run the test scenarios again. The expected result for tests 22-25 should now be True, this validates we are using the same formula for our rollout calculation.
```

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.sample.scenarios.md
```diff
@@ -0,0 +1,41 @@
+# Baseline Test Secenarios for Feature Management Libraries
+
+The following test scenarios use FeatureFlag.v2.0.0.schema.sample.json as the baseline file for testing the Feature Management libraries.
+
+| Test # | Feature Flag Name | Inputs                            | Expected Result | Description |
+|--------|-------------------|-----------------------------------|-----------------|-------------|
+| 1      | Alpha             | N/A                               | True            | An Enabled Feature Flag with no Filtered            |
+| 2      | Beta              | N/A                               | False           | A Disabled Feature Flag with no Filters             |
+| 3      | Gamma             | N/A                               | False           | Time Window filter where both Start and End have yet to happen            |
```
amerjusupovic, line 9:
```suggestion
| 3      | Gamma             | N/A                               | False           | Time Window filter where both Start and End have already passed           |
```

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.sample.json
```diff
@@ -0,0 +1,276 @@
...
+                                "Start": "Tue, 27 Jun 2023 06:00:00 GMT"
+                            }
+                        }
+                    ],
+                    "requirement_type": "Any"
+                }
+            },
+            {
+                "id": "Mu",
+                "description": "Same as Kappa, but requirement type is specified. Will always return true.",
```
amerjusupovic, line 205:
```suggestion
                "description": "Same as Lambda, but neither filter enables the flag. Will always return false.",
```

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.sample.json
```diff
@@ -0,0 +1,276 @@
...
+                                "End": "Tue, 27 Jun 2023 06:00:00 GMT"
+                            }
+                        }
+                    ],
+                    "requirement_type": "All"
+                }
+            },
+            {
+                "id": "Xi",
+                "description": "Requirement type All. Will always return false.",
```
amerjusupovic, line 252:
```suggestion
                "description": "Requirement type All. Will always return true.",
```

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.sample.json
zhenlan, line 1: It will be helpful if we can segment these samples, for example, vanilla feature flags, feature flags with filters, feature flags with variants, feature flags with everything, etc.  It will make it easier for customers to find the example they are looking for. The samples will also be more organized, so we know where to add when we have something new to add next time. Maybe you are already following this categorization. It's just not very clear at a glance.

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.sample.scenarios.md
```diff
@@ -0,0 +1,41 @@
...
+| 26     | Iota              | N/A                               | True            | Feature Flag with two feature filters, first returns true, so it's enabled.            |
+| 27     | Kappa             | N/A                               | True            | Feature Flag with two feature filters, second returns true, so it's enabled.            |
+| 28     | Lambda            | N/A                               | True            | Feature Flag with two feature filters and requirement type specified as Any. Second filter returns true            |
+| 29     | Mu                | N/A                               | False           | Feature Flag with two feature filters and requirement type specified as Any. Neither filter returns true            |
+| 30     | Nu                | user="Adam"                       | False           | Feature Flag with two feature filters and requirement type specified as All. Only the first filter returns true          |
+| 31     | Xi                | user="Adam"                       | True            | Feature Flag with two feature filters and requirement type specified as All. Both filters return true            |
```
zhiyuanliang-ms, line 37:
```suggestion
| Test # | Feature Flag Name | Inputs                            | Expected Result | Description                                                                                                               |
|--------|-------------------|-----------------------------------|-----------------|---------------------------------------------------------------------------------------------------------------------------|
| 1      | Alpha             | N/A                               | True            | An Enabled Feature Flag with no Filtered                                                                                  |
| 2      | Beta              | N/A                               | False           | A Disabled Feature Flag with no Filters                                                                                   |
| 3      | Gamma             | N/A                               | False           | Time Window filter where both Start and End have already passed                                                           |
| 4      | Delta             | N/A                               | False           | Time Window filter where neither Start nor End have happened                                                              |
| 5      | Epsilon           | N/A                               | True            | Time Window filter with a Start that has passed                                                                           |
| 6      | Zeta              | N/A                               | True            | Time Window filter where the End hasn't passed                                                                            |
| 7      | Eta               | user="Adam"                       | False           | Targeting Filter, Adam is not part of the default rollout                                                                 |
| 8      | Eta               | user="Ellie"                      | True            | Targeting Filter, Ellie is part of the default rollout                                                                    |
| 9      | Eta               | user="Alice"                      | True            | Targeting Filter, Alice is a targeted user                                                                                |
| 10     | Eta               | user="Adam" groups=["Stage1"]     | True            | Targeting Filter, Adam now part of the 100% rolled out group Stage1 is now targeted                                       |
| 11     | Eta               | groups=["Stage2"]                 | True            | Targeting Filter, no user Stage 2 is 50% rolloued out.                                                                    |
| 12     | Eta               | user="Adam" groups=["Stage2"]     | True            | Targeting Filter, Adam who is not part of the default rollout is part of the first 50% of Stage 2                         |
| 13     | Eta               | user="Chad" groups=["Stage2"]     | False           | Targeting Filter, Chad is neither part of the default rollout or part of the first 50% of Stage 2                         |
| 14     | Eta               | groups=["Stage3"]                 | False           | Targeting Filter, the Stage 3 is group is on the exclusion list                                                           |
| 15     | Eta               | user="Alice" groups=["Stage3"]    | False           | Targeting Filter, Alice who is targeted is excluded because she is part of the Stage 3 group                              |
| 16     | Eta               | user="Ellie" groups=["Stage3"]    | False           | Targeting Filter, Ellie who was enabled by the default rollout is now excluded as part of the Stage 3 group               |
| 17     | Eta               | user="Dave" groups=["Stage1"]     | False           | Targeting Filter, Dave is on the exclusion list, is still excluded even though he is part of the 100% rolled out Stage 1  |
| 18     | Theta             | user="Adam"                       | True            | Targeting Filter, 74% default rollout Adam is part of it.                                                                 |
| 19     | Theta             | user="Adam" groups=["Stage1"]     | True            | Targeting Filter, group is not part of default rollout calculation, no change                                             |
| 20     | Theta             | user="Adam" groups=["Stage2"]     | True            | Targeting Filter, group is not part of default rollout calculation, no change                                             |
| 21     | Theta             | user="Adam" groups=["Stage3"]     | True            | Targeting Filter, group is not part of default rollout calculation, no change                                             |
| 22     | Theta             | user="Brittney" groups=["Stage1"] | False           | Targeting Filter, 74% default rollout Brittney is not part of it                                                          |
| 23     | Theta             | user="Brittney" groups=["Stage2"] | False           | Targeting Filter, group is not part of default rollout calculation, no change                                             |
| 24     | Theta             | user="Brittney" groups=["Stage3"] | False           | Targeting Filter, group is not part of default rollout calculation, no change                                             |
| 25     | Theta             | user="Brittney" groups=["Stage4"] | False           | Targeting Filter, group is not part of default rollout calculation, no change                                             |
| 26     | Iota              | N/A                               | True            | Feature Flag with two feature filters, first returns true, so it's enabled.                                               |
| 27     | Kappa             | N/A                               | True            | Feature Flag with two feature filters, second returns true, so it's enabled.                                              |
| 28     | Lambda            | N/A                               | True            | Feature Flag with two feature filters and requirement type specified as Any. Second filter returns true                   |
| 29     | Mu                | N/A                               | False           | Feature Flag with two feature filters and requirement type specified as Any. Neither filter returns true                  |
| 30     | Nu                | user="Adam"                       | False           | Feature Flag with two feature filters and requirement type specified as All. Only the first filter returns true           |
| 31     | Xi                | user="Adam"                       | True            | Feature Flag with two feature filters and requirement type specified as All. Both filters return true                     |
```

This won't affect the appearance of the markdown after it is rendered, but this can make the raw file looks neat.

zhiyuanliang-ms:
FeatureFlagSchemaV2 contains the variants and telemetry. I think we should also cover variant allocation in the test suite.
Another important thing is that the targeting evaluation is sticky and it should give the same result on the same targeting context across different FM libs. The percentage allocation should use the same algorithm across all FM libs.

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.sample.json
```diff
@@ -0,0 +1,276 @@
...
+                                    }
+                                }
+                            }
+                        }
+                    ]
+                }
+            },
+            {
+                "id": "Theta",
+                "description": "A feature flag using a targeting filter, that will return true 74% of the time. Changing to 75% makes the user Brittney true.",
```
zhiyuanliang-ms, line 118: > Changing to 75% makes the user Brittney true.

We should include another testcase for it.

zhiyuanliang-ms:
Should we also include some invalid filter settings in the test suites?

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.sample.scenarios.md
Eskibear, line 1: I like this .md file which is clear and helpful. Some gut feelings, room to improve the user experience:
- Currently the granularity is per test. Might be more useful to group by feature, as I'm looking for how to define a feature.
- Might be clearer if sample.json files is also separated by scenarios, esp. when we add more and more samples in the future. And by using separated sample.json files, we can add a column, putting the links there. So I don't need to search the whole file for the feature. It can be convenient.
- How about adding a `Samples` folder, and use readme.md which will be auto rendered by GH when you open the folder.

E.g. What I imagine:
Folder structure, (might use consistent filename cases)
```
Samples
| - FeatureFlag.v2.0.0
    | - README.md
    | - NoFilters.sample.json
    | - TimeWindowFilter.sample.json
    | - TargetingFilter.sample.json
    | - RequirementType.sample.json
    | - ....

```

Each row of the table provides below information:
feature name, description, inputs (parameters), output (evaluation result), **corresponding sample.json file with link**

mrm9084:
> FeatureFlagSchemaV2 contains the variants and telemetry. I think we should also cover variant allocation in the test suite. Another important thing is that the targeting evaluation is sticky and it should give the same result on the same targeting context across different FM libs. The percentage targeting evaluation should use the same algorithm across all FM libs. And we should include some testcases to ensure that consistency.

I wrote this to validate Python, and it doesn't have variants yet, so they weren't included.

I wanted to get this up ASAP as we already found the Targeting issues.

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.sample.json
mrm9084: Can't we would need two feature flags with the same name to result in the same bucketing. We could have two different files, but that doesn't really work. My solution for now was the extra section. Maybe with breaking it out into sections like Zhenlan mentioned above it will work out better.

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.sample.scenarios.md
mrm9084: That might work, I was basing this off of the original email Zhenlan sent out. Probably can discuss online a bit before moving forward with this.

docs/FeatureManagement/Samples/FeatureFlag.v2.0.0.schema.sample.scenarios.md
zhenlan, line 1: Tables are nice, but I'm thinking if we should convert these tables to JSON format so our tests (in various languages) can consume them directly.

docs/FeatureManagement/Samples/FeatureFlag.v2.0.0.schema.sample.scenarios.md
```diff
@@ -0,0 +1,81 @@
+# Baseline Test Secenarios for Feature Management Libraries
+
+## No Filters Test Scenarios
+
+The following test scenarios use NoFilters.sample.json as the baseline file for testing the Feature Management libraries.
+
+| Test # | Feature Flag Name         | Expected Result | Description                                          |
+|--------|---------------------------|-----------------|------------------------------------------------------|
+| 1      | BooleanTrueFeatureFlag    | True            | An Enabled Feature Flag with no Filtered             |
```
zhenlan, line 9: NIT: ... with no _filters_.

docs/FeatureManagement/Samples/TargetingFilter.modified.sample.json
```diff
@@ -0,0 +1,60 @@
...
+                                    }
+                                }
+                            }
+                        }
+                    ]
+                }
+            },
+            {
+                "id": "RolloutPercentageUpdate",
+                "description": "A feature flag using a targeting filter, that will return true 74% of the time. Changing to 75% makes the user Brittney true.",
```
zhiyuanliang-ms, line 40: DefaultRolloutPercentage is 62 here. It seems like the description is wrong?

mrm9084:
Usage in python https://github.com/microsoft/FeatureManagement-Python/pull/20

docs/FeatureManagement/Samples/FeatureFlag.v2.0.0.schema.sample.scenarios.md
```diff
@@ -0,0 +1,36 @@
+# Baseline Test Secenarios for Feature Management Libraries
+
+## No Filters Test Scenarios
+
+The following test are the baseline file for testing the Feature Management libraries. These tests validate feature flags that have no filters applied to them.
+
+[No Filters Tests](NoFilters.tests.json)
+[No Filters Test Scenarios](NoFilters.sample.json)
```
zhenlan, line 8: Maybe we can add `- ` in front of each line so they won't be displayed in the same line by the md?

docs/FeatureManagement/Samples/NoFilters.tests.json
```diff
@@ -0,0 +1,42 @@
+{
+    "1": {
```
zhenlan, line 2: The index number is hard to maintain. Why is it necessary? Can we just use array? Everything in an array is automatically numbered.

docs/FeatureManagement/Samples/TargetingFilter.modified.tests.json
```diff
@@ -0,0 +1,50 @@
+{
+    "12": {
```
Eskibear, line 2: I'm also thinking why we don't directly use array. BTW, why here the index/key starts at "12"?

Eskibear: > So, the results of tests 12 through 19 should be True.

Got it. Probably it's because they are subset of the original test suites and you want to reference changed tests results by indices in description? I think we might be simple as below:
- Use the same full test cases, i.e. 1-19.
- Simply tell that Brittney is included after increasing rollout percentage, without mentioning specific tests cases.

docs/FeatureManagement/Samples/NoFilters.tests.json
mrm9084: I was wanting an ID field so could discuss an individual test, as we can't use the feature flag name to do that as it is many to one. We could change it to be something like `NoFilters1` or something like that. And/or we could move the value into the object.

docs/FeatureManagement/Samples/FeatureFlag.v2.0.0.schema.sample.scenarios.md
mrm9084: Added new lines so they don't show up on the same line.

docs/FeatureManagement/Samples/TargetingFilter.modified.tests.json
Eskibear: the update in f804375 LGTM

docs/FeatureManagement/Samples/NoFilters.tests.json
```diff
@@ -0,0 +1,50 @@
+[
+    {
+        "Test ID": "NoFilters1",
+        "Feature Flag Name": "BooleanTrueFeatureFlag",
```
Eskibear, line 4: `Feature Flag Name` vs `FeatureFlagName`, whether we need whitespaces in the keys? 

I assume that the xxx.sample.json and xxx.tests.json files will be used directly (probably via submodule) by FM libs of different languages to run unit tests, e.g. python, js, etc.
 
IMO without whitespaces would be more coding-friendly, while not harming the readability.

docs/FeatureManagement/Samples/FeatureFlag.v2.0.0.schema.sample.scenarios.md
```diff
@@ -0,0 +1,41 @@
+# Baseline Test Secenarios for Feature Management Libraries
+
+## No Filters Test Scenarios
+
+The following test are the baseline file for testing the Feature Management libraries. These tests validate feature flags that have no filters applied to them.
```
zhiyuanliang-ms, line 5:
```suggestion
The following tests are the baseline file for testing the Feature Management libraries. These tests validate feature flags that have no filters applied to them.
```

zhiyuanliang-ms: Same for the other section.

docs/FeatureManagement/Samples/FeatureFlag.v2.0.0.schema.sample.scenarios.md
```diff
@@ -0,0 +1,41 @@
...
+
+The following test are the baseline file for testing the Feature Management libraries. These tests validate that the same formula is being used to caluclate the rollout percentage. The modified files has a roolout percentage 1 higher than the original which takes Brittnety from being disabled to enabled. So, the results of tests 12 through 19 should be True.
+
+[Targeting Filter Tests](TargetingFilter.modified.tests.json)
+
+[Targeting Filter Test Scenarios](TargetingFilter.modified.sample.json)
+
+## Requirement Type Test Scenarios
+
+The following test are the baseline file for testing the Feature Management libraries. These tests validate feature flags that have requirement type field works with the Any and All operators.
```
zhiyuanliang-ms, line 37:
```suggestion
The following test are the baseline file for testing the Feature Management libraries. These tests validate feature flags that have requirement type field work with the Any and All operators.
```

docs/FeatureManagement/Samples/NoFilters.tests.json
```diff
@@ -0,0 +1,50 @@
...
+        "Expected Result": true,
+        "Description": "An Enabled Feature Flag with no Filters"
+    },
+    {
+        "Test ID": "NoFilters2",
+        "Feature Flag Name": "BooleanFalseFeatureFlag",
+        "Expected Result": false,
+        "Description": "A Disabled Feature Flag with no Filters"
```
zhiyuanliang-ms, line 12: It would be better to have description be a sentence with period.

docs/FeatureManagement/Samples/NoFilters.tests.json
```diff
@@ -0,0 +1,50 @@
...
+    {
+        "Test ID": "NoFilters1",
+        "Feature Flag Name": "BooleanTrueFeatureFlag",
+        "Expected Result": true,
+        "Description": "An Enabled Feature Flag with no Filters"
+    },
+    {
+        "Test ID": "NoFilters2",
+        "Feature Flag Name": "BooleanFalseFeatureFlag",
+        "Expected Result": false,
```
zhiyuanliang-ms, line 11: We have `Inputs` property in TargetingFilter.test. Would it be better to have an empty `Inputs` for all the other tests?

docs/FeatureManagement/Samples/NoFilters.sample.json
```diff
@@ -0,0 +1,60 @@
...
+            },
+            {
+                "id": "MinimalFeatureFlag",
+                "enabled": true
+            },
+            {
+                "id": "NoEnabled"
+            },
+            {
+                "id": "NoConditions",
```
zhiyuanliang-ms, line 52:
```suggestion
                "id": "EmptyConditions",
```

BTW, I think this has been covered in MinimalFeatureFlag.

docs/FeatureManagement/Samples/NoFilters.sample.json
```diff
@@ -0,0 +1,60 @@
...
+                "conditions": {
+                    "client_filters": []
+                }
+            },
+            {
+                "id": "MinimalFeatureFlag",
+                "enabled": true
+            },
+            {
+                "id": "NoEnabled"
```
zhiyuanliang-ms, line 49:
```suggestion
                "id": "NoEnabledFeatureFlag"
```

docs/FeatureManagement/Samples/RequirementType.sample.json
```diff
@@ -0,0 +1,144 @@
...
+                            "parameters": {
+                                "Start": "Thu, 29 Jun 2023 07:00:00 GMT",
+                                "End": "Wed, 30 Aug 2023 07:00:00 GMT"
+                            }
+                        }
+                    ]
+                }
+            },
+            {
+                "id": "SecondOrFeatureFlag",
```
zhiyuanliang-ms, line 27:
```suggestion
                "id": "DefaultRequirementTypeFeatureFlag2",
```

docs/FeatureManagement/Samples/RequirementType.sample.json
```diff
@@ -0,0 +1,144 @@
...
+                "description": "Same as Iota, but filter order is switched. Will always return true.",
```
zhiyuanliang-ms, line 28: Iota?

zhiyuanliang-ms: Same for the description of other feature flags.

docs/FeatureManagement/Samples/RequirementType.sample.json
```diff
@@ -0,0 +1,144 @@
...
+                            "name": "Microsoft.TimeWindow",
+                            "parameters": {
+                                "Start": "Tue, 27 Jun 2023 06:00:00 GMT"
+                            }
+                        }
+                    ]
+                }
+            },
+            {
+                "id": "FristRequirementTypeAnyFeatureFlag",
```
zhiyuanliang-ms, line 49:
```suggestion
                "id": "AnyRequirementTypeFeatureFlag1",
```

docs/FeatureManagement/Samples/RequirementType.sample.json
```diff
@@ -0,0 +1,144 @@
...
+                                "End": "Tue, 27 Jun 2023 06:00:00 GMT"
+                            }
+                        }
+                    ],
+                    "requirement_type": "Any"
+                }
+            },
+            {
+                "id": "TrueRequirementTypeAllFeatureFlag",
+                "description": "Requirement type All. Will always return false.",
```
zhiyuanliang-ms, line 96: The description doesn't match the id.

docs/FeatureManagement/Samples/RequirementType.sample.json
zhiyuanliang-ms: The reason to use AnyRequirementType instead of RequirementTypeAny is because I want to align with the previous "DefaultRequirementTypeFeatureFlag".

But I am ok with RequirementTypeAny

docs/FeatureManagement/Samples/RequirementType.sample.json
```diff
@@ -0,0 +1,144 @@
...
+                                "Start": "Tue, 27 Jun 2023 06:00:00 GMT"
+                            }
+                        }
+                    ],
+                    "requirement_type": "Any"
+                }
+            },
+            {
+                "id": "SecondtRequirementTypeAnyFeatureFlag",
+                "description": "Same as Kappa, but requirement type is specified. Will always return true.",
```
zhiyuanliang-ms, line 73: This feature flag will return false.

docs/FeatureManagement/Samples/NoFilters.sample.json
```diff
@@ -0,0 +1,60 @@
+{
+    "feature_management": {
+        "feature_flags": [
+            {
+                "id": "BooleanTrueFeatureFlag",
+                "description": "A simple feature flag, that always returns true.",
```
zhiyuanliang-ms, line 6: I am ok with saying "a feature flag returns true/false."
But I think it would be more appropriate to say "a feature flag is evaluated as enabled/activated/true".

docs/FeatureManagement/Samples/NoFilters.sample.json
```diff
@@ -0,0 +1,60 @@
...
+                "id": "StringFalseFeatureFlag",
+                "description": "A simple feature flag, that always returns true.",
+                "enabled": "false",
+                "conditions": {
+                    "client_filters": []
+                }
+            },
+            {
+                "id": "InvalidEnabledFeatureFlag",
+                "description": "A simple feature flag, that has an invalid enabled field.",
```
zhiyuanliang-ms, line 38:
```suggestion
                "description": "A simple feature flag, that has an invalid 'enabled' field.",
```

docs/FeatureManagement/Samples/RequirementType.sample.json
```diff
@@ -0,0 +1,144 @@
...
+                            "parameters": {
+                                "End": "Tue, 27 Jun 2023 06:00:00 GMT"
+                            }
+                        }
+                    ],
+                    "requirement_type": "Any"
+                }
+            },
+            {
+                "id": "TrueRequirementTypeAllFeatureFlag",
```
zhiyuanliang-ms, line 95: I suggest AllRequirementTypeFeatureFlag1.

docs/FeatureManagement/Samples/NoFilters.tests.json
```diff
@@ -0,0 +1,50 @@
...
+        "Description": "A Disabled Feature Flag with no Filters"
+    },
+    {
+        "Test ID": "NoFilters3",
+        "Feature Flag Name": "StringTrueFeatureFlag",
+        "Expected Result": true,
+        "Description": "An Enabled Feature Flag with no Filters"
```
zhiyuanliang-ms, line 18: The description doesn't match the testcase.

zhiyuanliang-ms:
I think we need to think carefully about the format of the testcase. Currently, we have a property called "Expected Result."

Now, this property is either true/false or "Exception."

If we want to introduce variants in the future, what should the expected result look like?

Once variants are introduced, we need to check both the variant and feature flag status.

Further, should we unify the exception which will be thrown in different FM libs? In this case, we can even specify the exception type in the testcases.


I proposed the following shape for our test suites:


```json
{
    "TestId": "ExampleTest-HelpfulTestNameForDebuging",
    "FeatureFlagName": "ExampleFeatureFlag",
    "Inputs": {}
    "ExpectedResult":{
       "FeatureStatus": boolean,
       "Variant": string,
       "Exception": string
    }
    "Description": "1. Describe the feature flag and the input. 2. Describe the reason why it should get the expected result.",
}

```

docs/FeatureManagement/Samples/TargetingFilter.sample.json
```diff
@@ -0,0 +1,60 @@
+{
+    "feature_management": {
+        "feature_flags": [
+            {
+                "id": "ComplexTargeting",
```
zhiyuanliang-ms, line 5: Let's name feature flags consistently.

How about ComplexTargetingFeatureFlag?

zhiyuanliang-ms, line 5: I suggest we add a separated feature flag for each targeting evaluation way.

docs/FeatureManagement/Samples/NoFilters.sample.json
mrm9084: I don't think so, it validates that the client_filters field is optional.

docs/FeatureManagement/Samples/NoFilters.sample.json
mrm9084: I think this is an open question in general. Do we or do we not include the word Feature Flag in the names? Can't it be assumed that they are all feature flags? I'm good with either way, but I was slightly leaning towards no including `FeatureFlag` in every feature flag name.

docs/FeatureManagement/Samples/RequirementType.sample.json
```diff
@@ -0,0 +1,144 @@
+{
+    "feature_management": {
+        "feature_flags": [
+            {
+                "id": "FirstOrFeatureFlag",
```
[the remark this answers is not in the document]
mrm9084: The `id` field in the samples file is the Feature Flag name, where the `Test ID` field in the tests file is the identifier, there isn't a 1 to 1 relationship. Some feature flags will have many tests, see targeting. They can't be the same. And Feature Flag names really shouldn't have numbers.

The idea with this flag name for example is that it is a feature flag with two items and is Any, the first of with returns true, so the name is FirstOr, where the next one is Second or as the second one returns true in that one.

docs/FeatureManagement/Samples/RequirementType.sample.json
mrm9084: I don't like `AnyRequirementType` as it reads as "any requirement type` not, "the value of the requirement type is any".

docs/FeatureManagement/Samples/RequirementType.sample.json
mrm9084: Yeah, I think somewhere in the process the values got switched. Updated.

docs/FeatureManagement/Samples/NoFilters.sample.json
mrm9084: I was trying to avoid the word evaluated as it sort of makes it sound like we look at filters. I swapped it to the description that was in the tests file.

docs/FeatureManagement/Samples/NoFilters.tests.json
mrm9084: Yeah, I've seen a few issues like this. I used co-pilot to move from the markdown tables to JSON, I don't think it worked as good as I hoped.

docs/FeatureManagement/Samples/TargetingFilter.sample.json
mrm9084: Started a thread on this above. Let's talk about it there https://github.com/Azure/AppConfiguration/pull/890#discussion_r1540636226.

mrm9084: I'm not sure what you mean by "evaluation way". Do you mean the different parameters? I'm fine with that, but we also need a full one, as we need to make sure everything works, not just individual parts.

At this point I don't want to add any more tests to this PR. Once it gets merged we can start to add more. The more tests added in this PR will just make it longer to approve.

docs/FeatureManagement/Samples/NoFilters.tests.json
```diff
@@ -0,0 +1,50 @@
+[
+    {
+        "TestID": "NoFilters1",
+        "FeatureFlagName": "BooleanTrueFeatureFlag",
+        "ExpectedResult": true,
+        "Description": "An Enabled Feature Flag with no Filters."
+    },
```
mrm9084, line 7: > I think we need to think carefully about the format of the testcase. Currently, we have a property called "Expected Result."
> 
> Now, this property is either true/false or "Exception."
> 
> If we want to introduce variants in the future, what should the expected result look like?
> 
> Once variants are introduced, we need to check both the variant and feature flag status.
> 
> Further, should we unify the exception which will be thrown in different FM libs? In this case, we can even specify the exception type in the testcases. After we release variant, there can be multiple kinds of exceptions. This can make sure our FM libs have the unified behavior.
> 
> Besides, we can add some testcases for invalid feature filter settings. For example, for time window filter, the start should be earlier than the end. This kind of validation for feature filter settings will become important after we release recurring time window. The setting will be much more complex.
> 
> I proposed the following shape for our test suites:
> 
> ```json
> {
>     "TestId": "ExampleTest-HelpfulNameForDebuging",
>     "FeatureFlagName": "ExampleFeatureFlag",
>     "Inputs": {}
>     "ExpectedResult":{
>        "FeatureStatus": true,
>        "Variant": "",
>        "Exception": ""
>     }
>     "Description": "1. Describe the feature flag and the input. 2. Describe the reason why it should get the expected result.",
> }
> ```
> 
> IMO, the test suite and samples we are implementing here is super valuable. It will be used by all FM libs in the future. Besides, the repo is open to our customer. Here can be the place for our customer to learn how to use feature management. We should put more efforts on it.

Good Point but, I think for "ExpectedResult"

{
  "IsEnabled": "true"/"false"/"exception" # This is a string
  "GetVariant": { # This would be "empty" for None/Null
    "name": "",
   "configuration": <This part may be tricky in some languages>
  }
}

Swapped the names to match the method names. Variants are more complex so it will need to be an object.

The exception part is tricky as we can say it will throw an exception, but the type will be different for each one. We could have the word exception be the value, but the message we throw could possibly be the same. Thoughts?

docs/FeatureManagement/Samples/TargetingFilter.sample.json
zhiyuanliang-ms: > Do you mean the different parameters?

Yes. Sure, we need to have a full one. I mean we can have feature flags: GroupTargetingFeatureFlag, UserTargetingFeatureFlag, DefaultRolloutFeatureFlag, ....

These flags are used to test the most simple and basic scenarios.

zhiyuanliang-ms: > At this point I don't want to add any more tests to this PR. Once it gets merged we can start to add more. The more tests added in this PR will just make it longer to approve.

If our plan is to keep working on this test suite and there will be more PRs for it, I agreed that we can keep it simple for now.

docs/FeatureManagement/Samples/NoFilters.sample.json
zhiyuanliang-ms: I am ok with not including `FeatureFlag` in every feature flag name. But we should make the name style consistent.

docs/FeatureManagement/Samples/RequirementType.sample.json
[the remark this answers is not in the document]
zhiyuanliang-ms: > The idea with this flag name for example is that it is a feature flag with two items and is Any, the first of with returns true, so the name is FirstOr, where the next one is Second or as the second one returns true in that one.

Now I get it. How about "DefaultRequirementTypeFirstFilterPassed" and "DefaultRequirementTypeLastFilterPassed"

docs/FeatureManagement/Samples/RequirementType.sample.json
zhiyuanliang-ms: Make sense.

docs/FeatureManagement/Samples/NoFilters.sample.json
zhiyuanliang-ms: The new descriptions look good to me.

docs/FeatureManagement/Samples/NoFilters.tests.json
zhiyuanliang-ms:
```
{
  "IsEnabled": "true"/"false"/"exception" # This is a string
  "GetVariant": { # This would be "empty" for None/Null
    "name": "",
    "configuration": <This part may be tricky in some languages>
  }
}
```
I like this one. 

> The exception part is tricky as we can say it will throw an exception, but the type will be different for each one. We could have the word exception be the value, but the message we throw could possibly be the same. Thoughts?

I just realized that exception can from either the `IsEnabled` or `GetVariant` method. So I think we'd better have a separated property for exception.

How about this?

```javascript
"exception":{
    "hasException": true/false // this can be removed, since empty "exception" means no exception
    "message/type": "xxx",
    "when/from":"xxx"
}
```

docs/FeatureManagement/Samples/TargetingFilter.sample.json
mrm9084: Think of this as V1, if you really look at this, we are mostly just doing positive test cases at the moment. I will probably add the negative test cases when I add variants.

docs/FeatureManagement/Samples/NoFilters.tests.json
mrm9084: I'm good with your suggestion, though what do you mean by "when/from", do you have an example.

docs/FeatureManagement/Samples/NoFilters.tests.json
```diff
@@ -0,0 +1,50 @@
...
+        "Description": "An Enabled Feature Flag with no Filters"
+    },
+    {
+        "Test ID": "NoFilters2",
+        "Feature Flag Name": "BooleanFalseFeatureFlag",
+        "Expected Result": false,
+        "Description": "A Disabled Feature Flag with no Filters"
+    },
+    {
+        "Test ID": "NoFilters3",
```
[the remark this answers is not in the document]
mrm9084: How about this for the format {Group/File Name}-{Filter Name}-{Optional Descriptor if more than one.}
For Example: "NoFilters-BooleanTrue" and "TargetingFilter-ComplexTargeting-DisabledDefautlRollout"

docs/FeatureManagement/Samples/TargetingFilter.modified.sample.json
```diff
@@ -0,0 +1,60 @@
...
+                                    }
+                                }
+                            }
+                        }
+                    ]
+                }
+            },
+            {
+                "id": "RolloutPercentageUpdate",
+                "description": "A feature flag using a targeting filter, that will return true 61% of the time. Changing to 62% makes the user Brittney true.",
```
amerjusupovic, line 40: I think there's some mismatches with description/test function for this flag, specifically with the Brittney scenario

docs/FeatureManagement/Samples/NoFilters.tests.json
zhiyuanliang-ms: "when/from" means which method this exception is from?

Does it happen in GetVariant or IsEnabled?

docs/FeatureManagement/Samples/TargetingFilter.modified.sample.json
```diff
@@ -0,0 +1,60 @@
...
+                                        "Groups": ["Stage3"]
+                                    }
+                                }
+                            }
+                        }
+                    ]
+                }
+            },
+            {
+                "id": "RolloutPercentageUpdate",
```
zhiyuanliang-ms, line 39: I am thinking this feature flag is not very easy to follow for those people who don't have the context. 
Currently, the ComplexTargeting flag in TargetingFilter.modified is not used. And for people have no context, they may feel strange why we have TargetingFilter and TargetingFilter.modified. And the testcases for Brittney locate in two places.

I still think, instead of having two feature flags both named "RolloutPercentageUpdate" in two files, it would be better to have two feature flags: "RolloutPercentage61" and "RolloutPercentage62" in one file. And we should put all testcases around them in one file.

docs/FeatureManagement/Samples/TargetingFilter.sample.json
```diff
@@ -0,0 +1,60 @@
...
+                                    }
+                                }
+                            }
+                        }
+                    ]
+                }
+            },
+            {
+                "id": "RolloutPercentageUpdate",
+                "description": "A feature flag using a targeting filter, that will return true 61% of the time. Changing to 62% makes the user Brittney true.",
```
zhiyuanliang-ms, line 40: Since we have talked about Brittney will hit the 62%, should we roughly explain the percentage targeting algorithm somewhere?

docs/FeatureManagement/Samples/TargetingFilter.tests.json
```diff
@@ -0,0 +1,211 @@
...
+    {
+        "TestID": "TargetingFilter-ComplexTargeting-ExcludedGroup",
+        "FeatureFlagName": "ComplexTargeting",
+        "Inputs": {"groups":["Stage3"]},
+        "ExpectedResult": {
+            "IsEnabled": "false",
+            "Variant": {}
+        },
+        "Exception": {},
+        "Description": "Targeting Filter, the Stage 3 is group is on the exclusion list."
```
zhiyuanliang-ms, line 88:
```suggestion
        "Description": "Targeting Filter, the Stage 3 is the group on the exclusion list."
```

docs/FeatureManagement/Samples/TargetingFilter.tests.json
```diff
@@ -0,0 +1,211 @@
...
+    {
+        "TestID": "TargetingFilter-ComplexTargeting-DisabledTargetedGroup50",
+        "FeatureFlagName": "ComplexTargeting",
+        "Inputs": {"user":"Chad", "groups":["Stage2"]},
+        "ExpectedResult": {
+            "IsEnabled": "false",
+            "Variant": {}
+        },
+        "Exception": {},
+        "Description": "Targeting Filter, Chad is neither part of the default rollout or part of the first 50% of Stage 2."
```
zhiyuanliang-ms, line 77:
```suggestion
        "Description": "Targeting Filter, Chad is neither part of the default rollout nor part of the first 50% of Stage 2."
```

docs/FeatureManagement/Samples/NoFilters.tests.json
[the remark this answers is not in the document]
zhiyuanliang-ms: Sounds good.

docs/FeatureManagement/Samples/TargetingFilter.tests.json
```diff
@@ -0,0 +1,211 @@
...
+    {
+        "TestID": "TargetingFilter-ComplexTargeting-TargetedGroup",
+        "FeatureFlagName": "ComplexTargeting",
+        "Inputs": {"user":"Aiden", "groups":["Stage1"]},
+        "ExpectedResult": {
+            "IsEnabled": "true",
+            "Variant": {}
+        },
+        "Exception": {},
+        "Description": "Targeting Filter, Aiden now part of the 100% rolled out group Stage1 is now targeted."
```
zhiyuanliang-ms, line 44: How about this: "Targeting Filter, Aiden is now targeted because Stage1 is 100% rolled out." ?

docs/FeatureManagement/Samples/TargetingFilter.tests.json
```diff
@@ -0,0 +1,211 @@
...
+    {
+        "TestID": "TargetingFilter-ComplexTargeting-DisabledTargetedGroup",
+        "FeatureFlagName": "ComplexTargeting",
+        "Inputs": {"groups":["Stage2"]},
+        "ExpectedResult": {
+            "IsEnabled": "false",
+            "Variant": {}
+        },
+        "Exception": {},
+        "Description": "Targeting Filter, no user Stage 2 is 50% rolloued out."
```
zhiyuanliang-ms, line 55: How about something like: empty/no user will hit the 50% rollout of group stage 2, so it is targeted.

docs/FeatureManagement/Samples/TargetingFilter.tests.json
```diff
@@ -0,0 +1,211 @@
...
+    {
+        "TestID": "TargetingFilter-ComplexTargeting-ExcludedGroupTargetedUser",
+        "FeatureFlagName": "ComplexTargeting",
+        "Inputs": {"user":"Alice", "groups":["Stage3"]},
+        "ExpectedResult": {
+            "IsEnabled": "false",
+            "Variant": {}
+        },
+        "Exception": {},
+        "Description": "Targeting Filter, Alice who is targeted is excluded because she is part of the Stage 3 group."
```
zhiyuanliang-ms, line 99: How about this: Alice is excluded because she is part of the Stage 3 group, even if she is an included user. Exclusion rules have higher priority than inclusion rules.

BTW, I feel like we should introduce the targeting evaluation flow somewhere . It can be a chart or comments. 
This will be difficult for people have no context about targeting evaluation to follow.

docs/FeatureManagement/Samples/TargetingFilter.tests.json
```diff
@@ -0,0 +1,211 @@
...
+        "Inputs": {"user":"Aiden", "groups":["Stage1"]},
+        "ExpectedResult": {
+            "IsEnabled": "true",
+            "Variant": {}
+        },
+        "Exception": {},
+        "Description": "Targeting Filter, group is not part of default rollout calculation, no change."
+    },
+    {
+        "TestID": "TargetingFilter-RolloutPercentageUpdate-AidenStage261",
```
zhiyuanliang-ms, line 146: AidenStage2-percentage61

Or even better, "TargetingFilter-RolloutPercentage61-AidenStage2" if we follow this: https://github.com/Azure/AppConfiguration/pull/890/files#r1545974176

docs/FeatureManagement/Samples/NoFilters.tests.json
mrm9084: > "when/from" means which method this exception is from?
> 
> Does it happen in GetVariant or IsEnabled?

The `is_enabled` and `get_variant` should have if it's an exception or not if we follow the above. Having her complicates things if both throw the exception.

docs/FeatureManagement/Samples/TargetingFilter.modified.sample.json
mrm9084: The idea is you don't need to understand these tests in order to validate the feature management libraries. And this is a really important check. It's validating that the algorithm we are using matches across all of the libraries.

We **CAN'T** change the name as that would change the targeting of the rollout and ruin the whole point of the test. The idea is that as 61% we have a user that isn't part of the rollout but at 62% they are. That will not be the case if the name of the feature flag changes.

docs/FeatureManagement/Samples/TargetingFilter.sample.json
mrm9084: Why? These are validation tests for our feature management libraries. Do we expect customer to implement there own versions of the libraries that matches our own?

docs/FeatureManagement/Samples/TargetingFilter.tests.json
mrm9084: These files are mainly validation files for our libraries. We shouldn't be explaining usage here, at most we redirect them to the actually libraries that already explain the logic.

zhenlan: I am all for making these samples and expected results as clear as possible for us and customers, but they are not replacement for documentation. I don't expect someone to learn feature management solely from these samples.

docs/FeatureManagement/Samples/NoFilters.sample.json
```diff
@@ -0,0 +1,60 @@
...
+            {
+                "id": "Minimal",
+                "enabled": true
+            },
+            {
+                "id": "NoEnabled"
+            },
+            {
+                "id": "EmptyConditions",
+                "description": "A Disabled Feature Flag with no Conditions.",
```
ivywei0125, line 53: Better to use "empty Conditions" instead of "no Conditions".
"no Conditions" sounds like there is no `conditions` property.

Eskibear: It's a little confusing by saying "a disabled feature flag" while in this sample it's `enabled: true`.  Maybe something like "A disabled feature flag **due to** empty conditions."?

BTW, there is chaos about capitalization, is that on purpose?

ivywei0125: I'm also confused with the capitalization.

docs/FeatureManagement/Samples/RequirementType.sample.json
```diff
@@ -0,0 +1,144 @@
...
+                                "Start": "Tue, 27 Jun 2023 06:00:00 GMT"
+                            }
+                        }
+                    ],
+                    "requirement_type": "Any"
+                }
+            },
+            {
+                "id": "RequirementTypeAllPassed",
+                "description": "Requirement type All. Will always return false.",
```
ivywei0125, line 96: For "RequirementTypeAllPassed" test case, it says `Will always return false.` in the description. Why it always return false?
The 1st and 2nd fiters should both return true?

docs/FeatureManagement/Samples/RequirementType.tests.json
```diff
@@ -0,0 +1,68 @@
...
+    {
+        "TestID": "RequirementType-RequirementTypeAllPassed",
+        "FeatureFlagName": "RequirementTypeAllPassed",
+        "Inputs": {"user":"Adam"},
+        "ExpectedResult": {
+            "IsEnabled": "true",
+            "Variant": {}
+        },
+        "Exception": {},
+        "Description": "Feature Flag with two feature filters and requirement type specified as All. Only the first filter returns true."
```
ivywei0125, line 55: Does this description match the test case? It should be "Both filters return true"

docs/FeatureManagement/Samples/RequirementType.tests.json
```diff
@@ -0,0 +1,68 @@
...
+    {
+        "TestID": "RequirementType-RequirementTypeAllLastFilterFailed",
+        "FeatureFlagName": "RequirementTypeAllLastFilterFailed",
+        "Inputs": {"user":"Adam"},
+        "ExpectedResult": {
+            "IsEnabled": "false",
+            "Variant": {}
+        },
+        "Exception": {},
+        "Description": "Feature Flag with two feature filters and requirement type specified as All. Both filters return true."
```
ivywei0125, line 66: Need to update the description. Should be "Only the first filter returns true."

docs/FeatureManagement/Samples/NoFilters.sample.json
zhenlan: Why is this a "disabled" feature flag?

docs/FeatureManagement/Samples/TargetingFilter.modified.tests.json
```diff
@@ -0,0 +1,90 @@
...
+    {
+        "TestID": "TargetingFilterModified-RolloutPercentageUpdate-Brittney62",
+        "FeatureFlagName": "RolloutPercentageUpdate",
+        "Inputs": {"user":"Brittney"},
+        "ExpectedResult": {
+            "IsEnabled": "true",
+            "Variant": {}
+        },
+        "Exception": {},
+        "Description": "Targeting Filter, 62% default rollout Brittney is not part of it."
```
ivywei0125, line 55: If Brittney is not part of it, shall we return "false" as the expected result?

docs/FeatureManagement/Samples/TargetingFilter.modified.tests.json
```diff
@@ -0,0 +1,90 @@
...
+        "Inputs": {"user":"Aiden"},
+        "ExpectedResult": {
+            "IsEnabled": "true",
+            "Variant": {}
+        },
+        "Exception": {},
+        "Description": "Targeting Filter, 62% default rollout Aiden is part of it."
+    },
+    {
+        "TestID": "TargetingFilterModified-RolloutPercentageUpdate-AidenStage162",
```
ivywei0125, line 14: Better to use the "Aiden-Stage1-62" instead of "AidenStage162".

docs/FeatureManagement/Samples/TargetingFilter.modified.tests.json
mrm9084: Copy Paste issue from the other description..

docs/FeatureManagement/Samples/RequirementType.tests.json
mrm9084: Seems like the descriptions on the two filters go switch somehow... updating.

docs/FeatureManagement/Samples/NoFilters.sample.json
mrm9084: Never mind just took a look at it. It should say enabled.

docs/FeatureManagement/Samples/TargetingFilter.tests.json
mrm9084: We can't take the linked suggestion, see my comment on it. The ID field is just to ID the test, if it fails. I just want them to be predictable. `{TestFile}-{FeatureFlagName}-{TestName}`. I also don't like changing the feature flag name on these targeting tests as it changes what the results are as the feature flag name is used to create the bucketing.

docs/FeatureManagement/Samples/TargetingFilter.modified.tests.json
mrm9084: I don't mind us changing these around a bit once we get everything merged in. I just would like this as a baseline so I can start expanding on the tests. We mainly have positive use cases here + Variants just got merged into python and I would like to add those too.

docs/FeatureManagement/Samples/TargetingFilter.sample.json
zhiyuanliang-ms: My concern is that our customer will also look these samples. This is a blackbox for them. They may feel confused why Brittney hits 62%. But I am now kind of agree with you. This is validation test. People are not expected to learn feature management solely from these samples.

docs/FeatureManagement/Samples/TargetingFilter.tests.json
ivywei0125: If we want to keep this format `{TestFile}-{FeatureFlagName}-{TestName}`, then we could use "TargetingFilter-RolloutPercentageUpdate-AidenStage2RolloutPercentage61"? It's quite long, but would be clearer.

docs/FeatureManagement/Samples/TargetingFilter.modified.tests.json
ivywei0125: >I don't mind us changing these around a bit once we get everything merged in. 
Do you mean we could merge this PR first, and then update the `TestID` property once the format is settled down?

mrm9084: sure

rossgrambo: I don't entirely see the reason for IDs. These are already identifiable as {filename}[index]. I'd rather have no id field and just have a friendly name for the test. Or no name at all.

docs/FeatureManagement/Samples/FeatureFlag.v2.0.0.schema.sample.scenarios.md
```diff
@@ -0,0 +1,41 @@
...
+
+The following tests are the baseline file for testing the Feature Management libraries. These tests validate feature flags that have targeting filters applied to them.
+
+[Targeting Filter Tests](TargetingFilter.tests.json)
+
+[Targeting Filter Test Scenarios](TargetingFilter.sample.json)
+
+### Validation of Rollout Percentage
+
+The following tests are the baseline file for testing the Feature Management libraries. These tests validate that the same formula is being used to caluclate the rollout percentage. The modified files has a roolout percentage 1 higher than the original which takes Brittnety from being disabled to enabled. So, the results of tests 12 through 19 should be True.
```
rossgrambo, line 29:
```suggestion
The following tests are the baseline file for testing the Feature Management libraries. These tests validate that the same formula is being used to calculate the rollout percentage. The modified files have a rollout percentage 1 higher than the original which takes Brittney from being disabled to enabled. So, the results of tests 12 through 19 should be True.
```

docs/FeatureManagement/Samples/TargetingFilter.modified.sample.json
```diff
@@ -0,0 +1,60 @@
+{
+    "feature_management": {
+        "feature_flags": [
+            {
+                "id": "ComplexTargeting",
```
rossgrambo, line 5: This flag is a duplicate and not used as far as I can tell.

docs/FeatureManagement/Samples/TargetingFilter.sample.json
```diff
@@ -0,0 +1,60 @@
+{
+    "feature_management": {
+        "feature_flags": [
+            {
+                "id": "ComplexTargeting",
+                "description": "A feature flag using a targeting filter, that will return true for Alice, Stage1, and 50% of Stage2, and false for Dave and Stage3. The default rollout percentage is 25%.",
```
rossgrambo, line 6:
```suggestion
                "description": "A feature flag using a targeting filter, that will return true for Alice, Stage1, and 50% of Stage2. Dave and Stage3 are excluded. The default rollout percentage is 25%.",
```

docs/FeatureManagement/Samples/TimeWindowFilter.sample.json
```diff
@@ -0,0 +1,84 @@
...
+                    "client_filters": [
+                        {
+                            "name": "Microsoft.TimeWindow",
+                            "parameters": {
+                                "End": "Sat, 28 Jun 3023 06:05:00 GMT"
+                            }
+                        }
+                    ]
+                }
+            }
```
rossgrambo, line 81: Could add an exception case here where the End is < Start.

docs/FeatureManagement/Samples/NoFilters.tests.json
zhiyuanliang-ms: What will the exception from `get_variant` look like if we follow the *above*?

What do you mean by *above*? This one?

``` json
{
  "IsEnabled": "true"/"false"/"exception" # This is a string
  "GetVariant": { # This would be "empty" for None/Null
    "name": "",
    "configuration": <This part may be tricky in some languages>
  }
}
```

docs/FeatureManagement/Samples/NoFilters.tests.json
```diff
@@ -0,0 +1,92 @@
...
+        "Exception": {},
+        "Description": "A Disabled Feature Flag with no Filters."
+    },
+    {
+        "TestID": "NoFilters-InvalidEnabled",
+        "FeatureFlagName": "InvalidEnabled",
+        "Inputs": {},
+        "ExpectedResult": {
+            "IsEnabled": "Exception",
+            "Variant": {}
```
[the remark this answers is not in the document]
zhiyuanliang-ms: I also have the same question. Tracking in this thread:
https://github.com/Azure/AppConfiguration/pull/890#discussion_r1542245787

docs/FeatureManagement/Samples/TargetingFilter.modified.tests.json
mrm9084: @rossgrambo what index? They aren't in a list. This is the ID we print out when the test fails.

docs/FeatureManagement/Samples/NoFilters.tests.json
```diff
@@ -0,0 +1,92 @@
+[
+    {
+        "TestID": "NoFilters-BooleanTrue",
+        "FeatureFlagName": "BooleanTrue",
+        "Inputs": {},
+        "ExpectedResult": {
+            "IsEnabled": "true",
+            "Variant": {}
+        },
+        "Exception": {},
```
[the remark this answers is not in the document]
mrm9084: If that is the case we should remove expected result and just have section called IsEnabled and Variant, with with the expected return value and possible exception message. Like:

"IsEnabled":{
    "Result": "true"
},
"Variant"{ 
   "Result": "Exception",
  "Message": "The error message."
}

Thoughts @rossgrambo, @zhiyuanliang-ms

docs/FeatureManagement/Samples/NoFilters.tests.json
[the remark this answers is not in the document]
mrm9084: @rossgrambo, variants haven't been added to these tests yet. I wanted to get this done before variants are added. This was just just trying to figure out the schema.

docs/FeatureManagement/Samples/TargetingFilter.modified.sample.json
mrm9084: The idea is that we run these same tests but slightly changed % values, we can add at test for this bumping up, but haven't gotten around to it yet. All of this is sort of the fact that this didn't start out like this.

docs/FeatureManagement/Samples/TimeWindowFilter.sample.json
mrm9084: It's on my todo list of things to add once this is merged in.

docs/FeatureManagement/Samples/NoFilters.tests.json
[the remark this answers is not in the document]
zhiyuanliang-ms: Sounds good.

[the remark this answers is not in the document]
rossgrambo: That works. I think I prefer

```json
"IsEnabled": {
    "Result": true,
    "Exception": "The error message"
},
"Variant": {
    "Result": {},
    "Exception": "The error message"
}
```

or

```json
"Result": {
    "IsEnabled": true,
    "Variant": {}
},
"Exception": {
    "IsEnabled": "The error message",
    "Variant": "The error message"
}
```

Where the field is only defined if it's expected. That way IsEnabled.Result can just be a bool and Variant.Result can just be an object.

docs/FeatureManagement/Samples/NoFilters.tests.json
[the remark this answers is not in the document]
rossgrambo: Makes sense. Still would be nice to decide what the schema will look like for them.

docs/FeatureManagement/Samples/TargetingFilter.modified.tests.json
rossgrambo: The top level is an array. The index would be it's place in the array. ie: `json.parse("tests.json")[index]`.

The reason I say this is because the ID is essentially a composite key we could deduce rather than manually define/maintain. IE:

```python
id = fileName + "." + FeatureFlagName + "." + friendlyName
```

But as you said- you want them unique, so:

```python
id = fileName + "." + index + "." + FeatureFlagName + "." + friendlyName
```

And this id doesn't need to actually exist, but just be print out when a test fails.

rossgrambo: Alternatively we could this for a unique id:
```python
uniqueId = fileName + "." + index
```

But when a test fails only use the friendly name. Maybe including ID if the friendly name isn't clear enough.
```python
message = "Exception in " + friendlyName + "(" + uniqueId + ")"
```

docs/FeatureManagement/Samples/NoFilters.tests.json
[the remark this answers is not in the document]
mrm9084: @ross, why would a result come back if there is an error message?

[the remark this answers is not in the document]
mrm9084: @rossgrambo, why would a result come back if there is an error message.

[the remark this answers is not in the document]
mrm9084: I like,

```json
"IsEnabled": {
    "Result": true,
    "Exception": "The error message"
},
"Variant": {
    "Result": {},
    "Exception": "The error message"
}
```

As it's a bit more flexible.

docs/FeatureManagement/Samples/TargetingFilter.tests.json
mrm9084: The Id field has been removed. See https://github.com/Azure/AppConfiguration/pull/890#discussion_r1547060699 for more details.

docs/FeatureManagement/Samples/NoFilters.tests.json
```diff
@@ -0,0 +1,90 @@
+[
+    {
+        "FeatureFlagName": "BooleanTrue",
+        "Inputs": {},
```
rossgrambo, line 4: nit: We could omit `Input` when it's empty just to reduce the size of this even more. But I do see some value in being explicit.

docs/FeatureManagement/Samples/TargetingFilter.modified.sample.json
```diff
@@ -0,0 +1,60 @@
+{
```
rossgrambo, line 1: I think we should remove the `.modified` files and use a flag shaped like this to confirm Brittney and Aiden within the `TargetingFilter.sample.json`:

```json
{
    "id": "RolloutPercentageUpdate",
    "description": "A feature flag using a targeting filter, that will return true 61% of the time. Changing to 62% makes the user Brittney true.",
    "enabled": true,
    "requirement_type": "All",
    "conditions": {
        "client_filters": [
            {
                "name": "Microsoft.Targeting",
                "parameters": {
                    "Audience": {
                        "DefaultRolloutPercentage": 61
                    }
                }
            },
            {
                "name": "Microsoft.Targeting",
                "parameters": {
                    "Audience": {
                        "DefaultRolloutPercentage": 62
                    }
                }
            }
        ]
    }
}
```

rossgrambo (approved):

docs/FeatureManagement/Samples/NoFilters.tests.json
Eskibear: > "IsEnabled": "true"/"false"/"exception" # This is a string

Sorry about the delayed response as I just get back to work. Maybe you've already agreed on this. But the "true"/"false" of string type looks a little bit confusing, esp. when the key name "IsEnabled" indicates it should be a boolean... 

If we need something to represent an exception, can it be `null`? Meaning that we cannot determine whether it's enabled or not because of the exception. I.e. "IsEnabled": true/false/null

AFAIU, the only benefit of using string would be the extensibility. Will there be more possible values for `IsEnabled`?

docs/FeatureManagement/Samples/TargetingFilter.tests.json
```diff
@@ -0,0 +1,230 @@
...
+            "Result": null
+        },
+        "Description": "Targeting Filter, Aiden who is not part of the default rollout is part of the first 50% of Stage 2."
+    },
+    {
+        "FriendlyName": "DisabledTargetedGroup50",
+        "FeatureFlagName": "ComplexTargeting",
+        "Inputs": {"user":"Chad", "groups":["Stage2"]},
+        "IsEnabled": {
+            "Result": "false"
```
Eskibear, line 79: Following dotnet, I implemented the same `isTargeted` in JS. And the result for this test case should be true. (if I impl it correctly)
"Chad\nComplexTargeting\nStage2": ~33.8% < 50%

Inviting @zhiyuanliang-ms to confirm...

zhiyuanliang-ms: I validated this in .NET FM. This result should be true.

zhiyuanliang-ms, line 79: This should be true.

zhiyuanliang-ms (changes requested):

docs/FeatureManagement/Samples/NoFilters.tests.json
mrm9084: I expect it will be required on most of them once we add variants.

docs/FeatureManagement/Samples/TargetingFilter.modified.sample.json
mrm9084: What would requirement type all validate here? It's the same as doing a single one at 62% true.

docs/FeatureManagement/Samples/TargetingFilter.tests.json
mrm9084: @eskibear @zhiyuanliang-ms, which version are you on? This should be false. You might have different results based on your CPU architecture in the released version.

Eskibear: Thanks, then we probably want to use another userId which is >= 50% and be evaluated as false. 
E.g. `Cad`?
 - "Cad\nComplexTargeting\nStage2": ~80.3%

mrm9084: ? I want this test to return false.
(Pdb) context_id
'Chad\nStage2\nComplexTargeting\nStage2'
(Pdb) percentage
51.995079324579585

mrm9084: > Thanks, then we probably want to use another userId which is >= 50% and be evaluated as false. E.g. `Cad`?
> 
> * "Cad\nComplexTargeting\nStage2": ~80.3%

The given inputs **should** return false.

zhiyuanliang-ms: I think the context id should be "Chad\nComplexTargeting\nStage2" ?

zhiyuanliang-ms: > which version are you on? This should be false. You might have different results based on your CPU architecture in the released version.

It used the little endianness, the same as python.

mrm9084: I think there is a bug in python. Hard catch if we didn't do this as we never checked the results in the past.

zhiyuanliang-ms (approved):

PR #895, opened 2024-04-02
Update Microsoft FeatureManagement schema

zhiyuanliang-ms:
Update the feature flag schema reference to v2

### DIFF
--- a/docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
+++ b/docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
@@ -5,7 +5,8 @@
   "type": "object",
   "title": "An Azure App Configuration Feature Declaration",
   "required": [
-    "id"
+    "id",
+    "enabled"
   ],
   "properties": {
     "id": {
--- a/docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
+++ b/docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
@@ -5,8 +5,7 @@
   "type": "object",
   "title": "An Azure App Configuration Feature Declaration",
   "required": [
-    "id",
-    "enabled"
+    "id"
   ],
   "properties": {
     "id": {
--- a/docs/FeatureManagement/FeatureManagement.v1.0.0.schema.json
+++ b/docs/FeatureManagement/FeatureManagement.v1.0.0.schema.json
@@ -23,7 +23,7 @@
                     "title": "Feature Flags",
                     "description": "Declares feature flags based on Microsoft Feature Flag schema.",
                     "items": {
-                        "$ref": "https://raw.githubusercontent.com/Azure/AppConfiguration/main/docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json"
+                        "$ref": "https://raw.githubusercontent.com/Azure/AppConfiguration/main/docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json"
                     }
                 }
             }
--- a/docs/FeatureManagement/FeatureManagement.v2.0.0.schema.json
+++ b/docs/FeatureManagement/FeatureManagement.v2.0.0.schema.json
@@ -1,7 +1,7 @@
 {
     "definitions": {},
     "$schema": "http://json-schema.org/draft-07/schema#",
-    "$id": "http://azconfig.io/schemas/FeatureManagement/v1.0.0/FeatureManagement.json",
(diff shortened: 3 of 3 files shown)


docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
```diff
@@ -5,7 +5,8 @@
   "type": "object",
   "title": "An Azure App Configuration Feature Declaration",
   "required": [
-    "id"
+    "id",
+    "enabled"
```
zhenlan, line 9: I think all FM libraries treat this as optional today. Are we going to take a breaking change?

docs/FeatureManagement/FeatureManagement.v1.0.0.schema.json
```diff
@@ -23,7 +23,7 @@
                     "title": "Feature Flags",
                     "description": "Declares feature flags based on Microsoft Feature Flag schema.",
                     "items": {
-                        "$ref": "https://raw.githubusercontent.com/Azure/AppConfiguration/main/docs/FeatureManagement/FeatureFlag.v1.1.0.schema.json"
+                        "$ref": "https://raw.githubusercontent.com/Azure/AppConfiguration/main/docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json"
```
zhenlan, line 26: I'm not sure this is the right way to do this. Should we create a FeatureManagement.**v2.0.0**.schema.json instead?

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
zhiyuanliang-ms: Our portal has already treated `enabled` as required. (Portal/AppConfigPortalExtension/AppConfigPortalExtension/Client/Shared/JsonSchemas.ts Line200)

I think this is a bug. A feature flag without `enabled` looks like an invalid use case. 

> Are we going to take a breaking change?

IMO, yes. Microsoft schema support is still under preview. It is a good chance to take breaking change.

docs/FeatureManagement/FeatureManagement.v1.0.0.schema.json
zhiyuanliang-ms: I am ok with create FM v2 schema. If we have FF v2.1 schema in the future, should we just update the reference in FM v2 schema or create a new FM v2.1 schema?

zhenlan: We shouldn't change the FF schema version in place, so it should be FM v2.1 schema.

docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json
zhenlan: Portal is just one of tools to create FF. FM libraries are the ones to enforce it.

I don't disagree with you. I would make the "enabled" property required if we started from scratch. Without FM libraries enforcing it, it doesn't matter what schema says. Here are two options.

**Option 1**
- Make "enabled" required in schema.
- .NET/Spring FM libraries continue to treat it as optional for compatibility.
- Python/JS FM libraries throw if the "enabled" property is absent.

**Option 2**
- Leave "enabled" as optional in schema.
- .NET/Spring FM libraries continue to treat it as optional.
- Python/JS FM libraries treat a feature flag as disabled if the "enabled" property is absent (the same behavior as .NET/Spring).

zhiyuanliang-ms: > I would make the "enabled" property required if we started from scratch. 

In this case, I think option 2 is fine.

@jimmyca15 @rossgrambo @Eskibear @mrm9084 What do you think?

rossgrambo: Enabled already defaults to false in this schema- so calling it required here has no impact unless we removed that as well. I prefer removing it here and leaving the default.

mrm9084: I agree with @rossgrambo

jimmyca15: enabled has a default value. It doesn't make sense to have a default value and be required.

I don't see any reason to make it required.

zhiyuanliang-ms: > enabled has a default value.

I missed that. In this case, `enabled` should be required.

zhenlan (approved):

PR #903, opened 2024-04-23
Add json schema for built-in feature filters

zhiyuanliang-ms:


docs/FeatureManagement/FeatureFlag.v1.0.0.schema.json
```diff
@@ -76,8 +76,8 @@
                 "description": "The name used to refer to and require a client filter.",
                 "default": "",
                 "examples": [
-                  "Percentage",
```
mrm9084, line 79: These are example names, not names of actual filters. I don't think we should propose in the schema the "Microsoft" name on filters. We wouldn't expect customers to use our prefix.

docs/FeatureManagement/TimeWindowFilter.v1.0.0.schema.json
```diff
@@ -0,0 +1,47 @@
...
+            "Wed, 1 May 2024 12:00:00 GMT",
+            "Wed, 1 May 2024 20:00:00 +0800"
+          ]
+        },
+        "End": {
+          "$id": "#/properties/parameters/properties/End",
+          "type": "string",
+          "title": "Time Window Filter End Parameter",
+          "examples": [
+            "Wed, 1 May 2024 12:00:00 GMT",
```
mrm9084, line 36: I know it's just an example, but the Start and end dates shouldn't be the same value. We can just updated to Thursday.

docs/FeatureManagement/TimeWindowFilter.v2.0.0.schema.json
```diff
@@ -0,0 +1,160 @@
...
+                      "Tuesday",
+                      "Wednesday",
+                      "Thursday",
+                      "Friday",
+                      "Saturday",
+                      "Sunday"
+                    ]
+                  }
+                },
+                "FirstDayOfWeek": {
```
ivywei0125, line 89: Just want to confirm, we didn't provide "Sunday" as default value?

zhiyuanliang-ms: Yes. I forgot to specify the default value. Added.

docs/FeatureManagement/TargetingFilter.v1.0.0.schema.json
zhenlan, line 1: Should we name the file name _Microsoft.Targeting.*_?

zhiyuanliang-ms: I am ok with naming the file with "Microsoft.xxx". We don't have Microsoft prefix for feature flag and feature management schema. Should we change them all?

zhenlan: We don't have _Microsoft_ prefixed to feature flag or feature management anywhere. It's just branding. For filters, _Microsoft.Targeting_ is the actual complete name users put in their feature flags and we expect from the code.

If we do change the name, it will be unclear this is about filters, so we probably can put them in a subfolder, eg., _FeatureManagement/FeatureFilters_.

docs/FeatureManagement/FeatureFilters/Microsoft.TimeWindow.v2.0.0.schema.json
```diff
@@ -0,0 +1,161 @@
...
+            "Thu, 2 May 2024 12:00:00 GMT",
+            "Thu, 2 May 2024 20:00:00 +0800"
+          ]
+        },
+        "Recurrence": {
+          "$id": "#/properties/parameters/properties/Recurrence",
+          "type": "object",
+          "title": "Time Window Filter Recurrence Parameter",
+          "properties": {
+            "Pattern": {
```
Eskibear, line 47: is `Pattern` required?

Eskibear: and `Range`? seems no required properties specified for `Recurrence`.

docs/FeatureManagement/FeatureFilters/Microsoft.TimeWindow.v2.0.0.schema.json
```diff
@@ -0,0 +1,165 @@
...
+                    "Weekly"
+                  ]
+                },
+                "Interval": {
+                  "$id": "#/properties/parameters/properties/Recurrence/properties/Pattern/properties/Interval",
+                  "type": "number",
+                  "title": "Time Window Filter Recurrence Pattern Interval Parameter",
+                  "default": 1
+                },
+                "DaysOfWeek": {
```
zhiyuanliang-ms, line 74: `DaysOfWeek` and `FirstDayOfWeek` only apply for `Weekly` pattern. Need to mention this in description or somewhere else in this schema.

ivywei0125 (approved):
LGTM

PR #920, opened 2024-05-22
Python Feature Management Release Notes

mrm9084:
Removed package section for now as it has yet to be released.

### DIFF
--- a/releaseNotes/PythonFeatureManagement.md
+++ b/releaseNotes/PythonFeatureManagement.md
@@ -1,0 +1,19 @@
+# featuremanagement
+
+[Source code][source_code] | [Samples][samples]
+
+## 1.0.0b1 - May 22, 2024
+
+Initial Beta Release of the Feature Management for Python
+
+### New Features
+
+* Loading of feature flags from a file, see [feature management schema]
+* Loading of feature flags from Azure App Configuration.
+* Checking if a feature is enabled.
+* Default feature filters: TimeWindowFilter, TargetingFilter.
+* Custom feature filters.
+
+[samples]: https://github.com/microsoft/FeatureManagement-Python/tree/main/samples
+[source_code]: https://github.com/microsoft/FeatureManagement-Python
+[feature management schema]:https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureManagement.v2.0.0.schema.json
--- a/releaseNotes/PythonFeatureManagement.md
+++ b/releaseNotes/PythonFeatureManagement.md
@@ -2,16 +2,15 @@
 
 [Source code][source_code] | [Samples][samples]
 
-## 1.0.0b1 - May 22, 2024
+## 1.0.0b1 - May 24, 2024
 
-Initial Beta Release of the Feature Management for Python
+### Enhancement
 
-### New Features
+Initial release of the feature management support for Python.
 
-* Loading of feature flags from a file, see [feature management schema].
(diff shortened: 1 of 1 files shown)


releaseNotes/PythonFeatureManagement.md
```diff
@@ -0,0 +1,19 @@
+# featuremanagement
+
+[Source code][source_code] | [Samples][samples]
+
+## 1.0.0b1 - May 22, 2024
+
+Initial Beta Release of the Feature Management for Python
+
+### New Features
```
zhenlan, line 9: Enhancement

(Please refer to the release notes guidelines)

releaseNotes/PythonFeatureManagement.md
```diff
@@ -0,0 +1,19 @@
...
+
+* Loading of feature flags from a file, see [feature management schema].
+* Loading of feature flags from Azure App Configuration provider.
+* Checking if a feature is enabled.
+* Default feature filters: TimeWindowFilter, TargetingFilter.
+* Custom feature filters.
```
zhenlan, line 15: Suggest below:

### Enhancement

Initial release of the feature management support for Python
- Support for loading feature flags from ... (dictionary?).
- Support for basic feature flags with boolean states.
- Support for feature filters including built-in filters `TimeWindowFilter` and `TargetingFilter`.

releaseNotes/PythonFeatureManagement.md
```diff
@@ -0,0 +1,19 @@
...
+[Source code][source_code] | [Samples][samples]
+
+## 1.0.0b1 - May 22, 2024
+
+Initial Beta Release of the Feature Management for Python
+
+### New Features
+
+* Loading of feature flags from a file, see [feature management schema].
+* Loading of feature flags from Azure App Configuration provider.
```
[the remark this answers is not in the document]
mrm9084: I guess the first item is wrong, as we support json, which has been loaded from a file into a dictionary.

As for the second item, I thought it would be good just as a callout that it is supported, though it will need 1.2.0 or later.

releaseNotes/PythonFeatureManagement.md
mrm9084: I guess I felt Enhancement was wrong, as the library doesn't exist.

releaseNotes/PythonFeatureManagement.md
mrm9084: Shouldn't we call out custom filters too?

zhenlan: All feature filters are custom filters. It's just that we ship some "custom filters" out of the box as well.

releaseNotes/PythonFeatureManagement.md
zhenlan: I don't disagree. It's enhanced from nothing :). 

The reason for the guideline is for 1) consistency 2) saving all our time on debating.

releaseNotes/PythonFeatureManagement.md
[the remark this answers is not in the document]
zhenlan: The FM library doesn't read files, does it?

>  it will need 1.2.0 or later

This is something worth calling out. How about something like this?

Initial release of the feature management support for Python. Note that, version 1.2.0 or later of xxx (provider package name) is required for loading feature flags from Azure App Configuration.

zhenlan (approved):

PR #922, opened 2024-05-24
Python Provider 1.2.0 Release Notes

mrm9084:


releaseNotes/AzureAppConfigurationProviderPython.md
```diff
@@ -2,6 +2,19 @@
...
 [Source code][source_code] | [Package (Pypi)][package] | [Samples][samples]
 
+## 1.2.0 - May 23, 2024
+
+### Features
+
+* Added support for the App Configuration Feature Management V2.0.0 schema. This allows for the loading of feature flags from an App Configuration provider. For more information see [here](https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureManagement.v2.0.0.schema.json)
+  * Added `feature_flags_enabled` method to the `load` method, to enable loading of feature flags, by default when enabled all feature flags with no label are loaded.
+  * Added `feature_flag_selectors` which works similar to `SettingSelector` but for feature flags.
+  * Added `feature_flag_refresh_enabled` to the `load` method, to enable the refreshing of feature flags. Feature flags are refreshed with the same refresh method as configuration settings, though they are refreshed separatetly, so a change in a feature flag will not cause a refresh of configuration settings.
```
zhenlan, line 12: Given this is the first time we added the support for feature flags, I don't think we need to go into details of implementation too much. We should have documentation for this. BTW, it's not *App Configuration Feature Management schema* either (it's *Microsoft ...*). Anyways, that's also an implementation detail. The python provider doesn't support any other schema before. I would simplify all above to:

Added support for loading feature flags from Azure App Configuration.

releaseNotes/AzureAppConfigurationProviderPython.md
```diff
@@ -2,6 +2,19 @@
...
+
+### Bug Fixes
+
+* Fixed an bug where loading configuration settings could be slower to return than expected.
```
zhenlan, line 16: So it's a perf improvement and it doesn't have any functional impact? If so, this can be put under *Enhancement*.

> slower to return than expected

It's more useful to customers if we can explain the loading time is improved for what scenarios.

mrm9084: How is the updated. I'm not sure how much to put as we didn't have some ICM. I just noticed it during some load testing and got it updated.

releaseNotes/AzureAppConfigurationProviderPython.md
zhenlan: Yes.

If docs are ready, we will include the links to the release notes. Otherwise, it's fine. It' more important to get release notes out sooner.

releaseNotes/AzureAppConfigurationProviderPython.md
zhenlan: We don't need an ICM to fix an issue :) You decide if this is worth calling out based on its impact (not ICM).

mrm9084: I meant there is no issue to link to for it.

releaseNotes/AzureAppConfigurationProviderPython.md
```diff
@@ -2,6 +2,16 @@
...
 
+## 1.2.0 - May 24, 2024
+
+### Features
+
+* Added support for loading feature flags from Azure App Configuration.
+
+### Enhancement
+
+* Updated the reterival of configuration settings from provider to be more efficient.
```
zhenlan, line 13: "more efficient" can be confusing. Is it using less memory?

Improved the performance of configuration loading time.

mrm9084: Actually, it would be, don't have numbers. Only noticed the issue when large key value sets were selected, like loading thousands of keys.

But updated to your comment.

zhenlan: I thought it's time only based on your original note (_slower to return than expected_). If it's the overall perf, we can tweak the notes (based on what you said):

Improved the performance of loading a large set of key-values.

mrm9084: It technically did make loading a single key faster. Just that it's such a small amount that it isn't noticeable as an issue to large key counts.

Also, it was both taking up extra memory and taking longer by how the keys were being returned.

zhenlan: Okay. Based on what you said:

Improved the performance of data loading from Azure App Configuration, especially when you have a large set of key-values.

zhenlan (approved):

PR #938, opened 2024-07-11
Add linting rule

zhiyuanliang-ms:
## Why this PR?

Have a centralized copy of the .editorconfig file. I will also write a onenote page for it. But since our open-source client libraries also follow the linting rule set. I think here is a good place for it.

### DIFF
--- a/docs/Lint/charp/.editorconfig
+++ b/docs/Lint/charp/.editorconfig
@@ -1,0 +1,39 @@
+# editorconfig.org
+
+# top-most EditorConfig file
+root = true
+
+## Default settings ##
+[*]
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+
+## Formatting rule ##
+# https://learn.microsoft.com/dotnet/fundamentals/code-analysis/style-rules/ide0055
+dotnet_diagnostic.IDE0055.severity = error
+
+# 'Using' directive preferences
+dotnet_sort_system_directives_first = false
+
+# New line preferences
+dotnet_diagnostic.IDE2002.severity = error
+csharp_style_allow_blank_lines_between_consecutive_braces_experimental = false
+dotnet_diagnostic.IDE2004.severity = error
+csharp_style_allow_blank_line_after_colon_in_constructor_initializer_experimental = false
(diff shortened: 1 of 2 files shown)


zhenlan:
Who is the audience for these files? How are these files going to be used?

zhiyuanliang-ms:
> Who is the audience for these files? How are these files going to be used?

App Config team and anyone wants to contribute to our open-source client libraries would be the audience.

I think we should have a centralized copy of the linting rule set (.editorconfig file) to be referenced and ensure every repo will use the same rule set. I do create an onenote page for it. But we also have feature management and .net provider which are on github. So I think this public repo is a better place to keep the linting rule set.

juniwang:
Keeping a centralized copy as single source of truth would be a good idea. At the same time, it also brings in some sort of cross repo dependencies making it less convenient in setting up CI/CD(especially on ADO) and local env.

docs/Lint/charp/.editorconfig
```diff
@@ -0,0 +1,39 @@
+# editorconfig.org
```
rossgrambo, line 1: path should be `csharp`* instead of `charp`

rossgrambo:
I like to think this is the .editorconfig for our team that we agreed on. Each repo might have slight variations (if there's reason to)- but this is the team's / would be used for a new repo.

Although I don't think I personally prefer for it to be public. It's not necessarily bad- but I don't see clear value.

rossgrambo:
If we believe all `.editorconfig` files for our repos need to be tightly coupled / the same- I don't see value in having a copy here. I don't believe that's the case.

zhenlan:
I'm with @rossgrambo on this. This repo is for customers, not centralized storage for our team. External developers who may contribute to our components will follow/obey the linting rules of respective repos.

docs/Lint/charp/.editorconfig
zhiyuanliang-ms: Good catch! My bad. Fixed.

zhiyuanliang-ms:
@zhenlan 

> This repo is for customers

I agree with this. Then I will move the rule set copy to our onenote.

PR #944, opened 2024-07-24
Add basic apps for quickstarts in portal

wuthomas-mit:
Added JavaScript, Python, and Java Spring basic apps.


examples/QuickStart/JavaSpring/QuickStart/.gitignore
```diff
@@ -0,0 +1,33 @@
+HELP.md
```
zhiyuanliang-ms, line 1: Why does JavaSpring folded have a .gitignore file? Other languages don't have it.

examples/QuickStart/JavaSpring/QuickStart/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+			<artifactId>spring-cloud-azure-appconfiguration-config-web</artifactId>
+		</dependency>
+	</dependencies>
+
+	<dependencyManagement>
+		<dependencies>
+			<dependency>
+			<groupId>com.azure.spring</groupId>
+			<artifactId>spring-cloud-azure-dependencies</artifactId>
+			<version>5.8.0</version>
```
ivywei0125, line 55: Latest version is 5.14.0. Maybe we could use the latest version.

examples/QuickStart/JavaSpring/QuickStart/src/main/java/app/config/QuickStart/HelloController.java
```diff
@@ -0,0 +1,17 @@
...
+
+    public HelloController(MyProperties properties) {
+        this.properties = properties;
+    }
+
+    @GetMapping
+    public String getMessage() {
+        return "Message: " + properties.getMessage();
+    }
+}
```
mrm9084, line 17: Format should have a blank line at the end of the file.

examples/QuickStart/JavaSpring/QuickStart/src/main/resources/application.properties
```diff
@@ -0,0 +1 @@
+spring.application.name=QuickStart
```
mrm9084, line 1: This file isn't needed, this config can be moved to bootstrap.properties.

examples/QuickStart/JavaSpring/QuickStart/src/test/java/app/config/QuickStart/QuickStartApplicationTests.java
```diff
@@ -0,0 +1,13 @@
+package app.config.QuickStart;
+
+import org.junit.jupiter.api.Test;
+import org.springframework.boot.test.context.SpringBootTest;
+
+@SpringBootTest(properties = "spring.cloud.azure.appconfiguration.enabled=false")
```
mrm9084, line 6: Is there a point in having the test files if they are just disabled?

examples/QuickStart/Python/QuickStartApp/QuickStartApp.py
```diff
@@ -0,0 +1,20 @@
...
+
+# Find the key "message" and print its value.
+print(config["message"])
+
+# Connect to Azure App Configuration using SettingSelector.
+selects = {SettingSelector(key_filter="message*", label_filter="\0")}
+config = load(connection_string=connection_string, selects=selects)
+
+# Print True or False to indicate if "message" is found in Azure App Configuration.
+print("message found: " + str("message" in config))
```
mrm9084, line 20: Is there a point in showing this? The default label is `\0` so that isn't needed. You could make a slightly more complex sample with using trimming.

mrm9084 (changes requested):

examples/QuickStart/DotNet/QuickStartApp/Program.cs
```diff
@@ -3,7 +3,7 @@
 var builder = new ConfigurationBuilder();
 builder.AddAzureAppConfiguration(options =>
 {
-    options.Connect(Environment.GetEnvironmentVariable("ConnectionString"))
```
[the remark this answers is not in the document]
avanigupta: This app will be added to the QuickStart blade in portal. We want it to be as simple as possible. This avoids taking a dependency on Entra ID and makes it easy to get started with a fully functional app within a few minutes.
Its also easier to align the quickstarts with the existing .NET version: https://github.com/Azure/AppConfiguration/blob/main/examples/QuickStart/DotNet/QuickStartApp/Program.cs

examples/QuickStart/Python/QuickStartApp/QuickStartApp.py
avanigupta: Same here - we're adding a simple app which matches the functionality of existing .NET app

examples/QuickStart/DotNet/QuickStartApp/Program.cs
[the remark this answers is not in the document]
mrm9084: Isn't it in our timeline somewhere to disable connection by default?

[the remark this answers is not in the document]
mrm9084: Also, the `DefaultAzureCredential` can use the azure portal to login in most cases.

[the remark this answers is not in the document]
avanigupta: It depends on where the app is running. The main hinderance with EntraID in quick starts is that users need to assign data role and wait ~15 minutes before accessing the data.

examples/QuickStart/JavaScript/QuickStartApp/QuickStartApp.js
```diff
@@ -0,0 +1,21 @@
+const { load } = require("@azure/app-configuration-provider");
```
Eskibear, line 1: This sample looks good. Note:
Node.js runtime is required, and install the required dependencies via below command.
```bash
npm install @azure/app-configuration-provider
```

avanigupta: This would be part of the quickstart instructions like this:

image

examples/QuickStart/DotNet/QuickStartApp/Program.cs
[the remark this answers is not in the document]
mrm9084: This can be ignored for now.

examples/QuickStart/JavaSpring/QuickStart/src/main/resources/bootstrap.properties
```diff
@@ -0,0 +1,2 @@
+spring.application.name=QuickStart
+spring.cloud.azure.appconfiguration.stores[0].connection-string= ${APP_CONFIGURATION_CONNECTION_STRING}
```
mrm9084, line 2: NIT: missing new line at the end of the file

examples/QuickStart/Python/QuickStartApp/QuickStartApp.py
```diff
@@ -0,0 +1,13 @@
...
+)
+import os
+
+connection_string = os.environ.get("AZURE_APPCONFIG_CONNECTION_STRING")
+
+# Connect to Azure App Configuration using a connection string.
+config = load(connection_string=connection_string)
+
+# Find the key "message" and print its value.
+print(config["message"])
```
mrm9084, line 13: NIT: missing new line at the end of the file

mrm9084 (changes requested):

examples/QuickStart/JavaSpring/QuickStart/src/main/java/app/config/QuickStart/MyProperties.java
```diff
@@ -0,0 +1,15 @@
+package app.config.QuickStart;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "config")
```
mrm9084, line 4: We are thinking we should add some code comments.

```suggestion

// The `prefix` value of `@ConfigurationProperties` matches the created key of `/application/config.message` so all properties with that prefix are added to this `@ConfigurationProperties`
// Note: By default, `/application/` is removed from the key name.
@ConfigurationProperties(prefix = "config")
```

examples/QuickStart/JavaSpring/QuickStart/src/main/resources/bootstrap.properties
mrm9084, line 2:
```suggestion
# Either a connection string or endpoint needs to be provider per store.
# All possible configurations can be found in the [README](https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/spring/spring-cloud-azure-starter-appconfiguration-config)
spring.cloud.azure.appconfiguration.stores[0].connection-string= ${APP_CONFIGURATION_CONNECTION_STRING}
```

avanigupta (approved):

mrm9084 (approved):

PR #960, opened 2024-09-04
Feature Management repo link

mrm9084:
NOTE: Need to make Feature Management public before merging. Marking as draft till then.

Removing schema/samples linking to [Feature Management](https://github.com/microsoft/FeatureManagement) repo.


docs/FeatureManagement/readme.md
```diff
@@ -2,12 +2,4 @@
 
 Azure App Configuration provides feature management support through the creation of **Feature Flags**. Feature flags consist of a unique identifier and an enabled state that can be evaluated dynamically. These flags are stored in Azure App Configuration as key-values with a common set of conventions so that they can be identified among other non feature flag key-values.
```
zhenlan, line 3: The feature flags are more than enable/disable now. How about this?

Azure App Configuration provides feature management support through the creation of **Feature Flags**. These flags are stored in Azure App Configuration as key-values with a common set of conventions so that they can be identified among other non-feature flag key-values.

For more information, see [Feature Management](https://github.com/microsoft/FeatureManagement).

PR #973, opened 2024-09-26
Release note for JS FM 1.0.0 stable & 2.0.0 preview

zhiyuanliang-ms:


releaseNotes/JavaScriptFeatureManagement.md
```diff
@@ -0,0 +1,29 @@
...
+### Enhancements
+
+* The targeting filter can work properly in non-Node.js environments. [#25](https://github.com/microsoft/FeatureManagement-JavaScript/pull/25)
+
+## 1.0.0-preview.1 - May 17, 2024
+
+Added support for the following features.
+- Loading of feature flags from object and map.
+- Feature flag status evaluation with or without context.
+- Feature filters including built-in filters [`Microsoft.TimeWindow`](https://github.com/microsoft/FeatureManagement/blob/main/Schema/FeatureFilters/Microsoft.TimeWindow.v1.0.0.schema.json) and [`Microsoft.Targeting`](https://github.com/microsoft/FeatureManagement/blob/main/Schema/FeatureFilters/Microsoft.Targeting.v1.0.0.schema.json).
```
zhenlan, line 25: Built-in feature filters [`Microsoft.TimeWindow`](https://github.com/microsoft/FeatureManagement/blob/main/Schema/FeatureFilters/Microsoft.TimeWindow.v1.0.0.schema.json) and [`Microsoft.Targeting`](https://github.com/microsoft/FeatureManagement/blob/main/Schema/FeatureFilters/Microsoft.Targeting.v1.0.0.schema.json).

releaseNotes/JavaScriptFeatureManagement.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Enhancements
+
+* The targeting filter can work properly in non-Node.js environments. [#25](https://github.com/microsoft/FeatureManagement-JavaScript/pull/25)
+
+## 1.0.0-preview.1 - May 17, 2024
+
+Added support for the following features.
+- Loading of feature flags from object and map.
+- Feature flag status evaluation with or without context.
```
zhenlan, line 24: Is this what you mean?

Feature flag evaluation with ambient context or user-provided context.

releaseNotes/JavaScriptFeatureManagement.md
```diff
@@ -0,0 +1,29 @@
...
+## 1.0.0 - Sep 26, 2024
+
+This is the first stable release of the following features.
+- Loading of feature flags from object and map.
+- Feature flag status evaluation with or without context.
+- Feature filters including built-in filters [`Microsoft.TimeWindow`](https://github.com/microsoft/FeatureManagement/blob/main/Schema/FeatureFilters/Microsoft.TimeWindow.v1.0.0.schema.json) and [`Microsoft.Targeting`](https://github.com/microsoft/FeatureManagement/blob/main/Schema/FeatureFilters/Microsoft.Targeting.v1.0.0.schema.json).
+
+### Breaking Changes
+
+* The feature flag provider handles duplicate flags by using the last flag instead of the first. [#40](https://github.com/microsoft/FeatureManagement-JavaScript/pull/40)
```
zhenlan, line 14: This is the first stable release, so there is not much in terms of breaking change. We can add this to the section above as a feature.

releaseNotes/JavaScriptFeatureManagement.md
zhiyuanliang-ms: No. 

My original intention is to say you can call isEnabled("MyFeature") or isEnabled("MyFeature", context) to evaluate a feature flag, depending on whether the feature flag uses any contextual feature filter.

I don't mean to call out ambient context. Because we have made the library-level decision for targeting context design that we require users to always provide targeting context. I used to think for any contextual feature filter in JS FM, if you want to use any context for the filter, you should pass it as the parameter of the isEnabled call.

I just realized that the time window filter was actually using the "ambient context" (or do we consider the current time as "ambient context"?). In .NET FM, we have IContextualFilter and IFeatureFilter. But I just releazed all feature filters need some kind of context to do the evaluation. So the context must be ambient or provided as a parameter.

So now, I think your suggestion is good. `Feature flag evaluation with ambient context or user-provided context.`

releaseNotes/JavaScriptFeatureManagement.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Enhancements
+
+* The targeting filter can work properly in non-Node.js environments. [#25](https://github.com/microsoft/FeatureManagement-JavaScript/pull/25)
```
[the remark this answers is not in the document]
zhiyuanliang-ms: I don't consider it as a bug. Previously, I confirmed with Yan, the JS FM only targets on the Node.js scenario. That's why we wrote Node.js LTS version in the library prerequisites.

releaseNotes/JavaScriptFeatureManagement.md
```diff
@@ -1,7 +1,15 @@
-# Microsoft Feature Management for JavaScript
+# Microsoft Feature Management Application Insights for Browser
```
zhenlan, line 1: Changed in the wrong place?

releaseNotes/JavaScriptFeatureManagementApplicationInsightsBrowser.md
```diff
@@ -0,0 +1,13 @@
+# Microsoft Feature Management for JavaScript
```
zhenlan, line 1: How do you like this?

Microsoft JavaScript Feature Management for Application Insights in Browsers

releaseNotes/JavaScriptFeatureManagementApplicationInsightsBrowser.md
```diff
@@ -0,0 +1,13 @@
...
+
+[Source code][source_code] | [Package (npm)][package] | [Samples][samples]
+
+## 2.0.0-preview.1 - Oct 15, 2024
+
```
zhenlan, line 6: Enhancements

releaseNotes/JavaScriptFeatureManagement.md
```diff
@@ -1,7 +1,15 @@
...
 
 [Source code][source_code] | [Package (npm)][package] | [Samples][samples]
 
+## 2.0.0-preview.1 - Oct 15, 2024
+
+### Enhancement
```
zhenlan, line 7: Enhancement**s**

releaseNotes/JavaScriptFeatureManagementApplicationInsightsBrowser.md
```diff
@@ -0,0 +1,13 @@
...
+* Added a built-in method to create telemetry publisher with Application Insights.
```
zhenlan, line 7: Added support for publishing telemetry to Application Insights.

releaseNotes/JavaScriptFeatureManagementApplicationInsightsBrowser.md
```diff
@@ -0,0 +1,13 @@
...
+
+* Added a built-in `TrackEvent` method to send custom event to Application Insights with targeting id attached.
```
zhenlan, line 9: Does this make sense?

Introduced a `TrackEvent` API, allowing users to replace existing Application Insights `TrackEvent` calls to include targeting information in custom events sent to Application Insights.

releaseNotes/JavaScriptFeatureManagementApplicationInsightsBrowser.md
zhiyuanliang-ms: This is the initial release. So I didn't add it. Now added.

releaseNotes/JavaScriptFeatureManagementApplicationInsightsBrowser.md
zhiyuanliang-ms: Our function signature is different from the Application Insights' `TrackEvent`. 
appInsightsClient.TrackEvent("EventName") vs TrackEvent(appInsightsClient, targetingId, "EventName")

So my concern is that whether it will be misleading.

releaseNotes/JavaScriptFeatureManagement.md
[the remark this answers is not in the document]
zhenlan: But this is the first stable release of the package. There is no previous stable release to enhance upon. These are all just changes between preview releases. I'd suggest we remove it. The release notes are for customers. They are not change logs. This note adds more confusion than value to customers.

releaseNotes/JavaScriptFeatureManagementApplicationInsightsBrowser.md
zhenlan: Isn't the reason you name the new API `TrackEvent` that you want users to use the new API to replace the existing `TrackEvent`? We didn't say they have the same signature.

zhiyuanliang-ms: Ok, then it wil be fine.

releaseNotes/JavaScriptFeatureManagement.md
```diff
@@ -0,0 +1,34 @@
...
+
+* Added support for telemetry in feature flags. Telemetry is a powerful feature that allows you to track how your feature flags are being used. It provides insights into the effectiveness of your feature flags and helps you make data-driven decisions. Telemetry is particularly useful for feature experimentation (also known as A/B testing) and feature rollouts.
+
+## 1.0.0 - Sep 26, 2024
+
+This is the first stable release of the following features.
+- Loading of feature flags from object and map.
+- Feature flag evaluation with ambient context or user-provided context.
+- Built-in feature filters [`Microsoft.TimeWindow`](https://github.com/microsoft/FeatureManagement/blob/main/Schema/FeatureFilters/Microsoft.TimeWindow.v1.0.0.schema.json) and [`Microsoft.Targeting`](https://github.com/microsoft/FeatureManagement/blob/main/Schema/FeatureFilters/Microsoft.Targeting.v1.0.0.schema.json).
+- The feature flag provider handles duplicate flags by using the last flag instead of the first. [#40](https://github.com/microsoft/FeatureManagement-JavaScript/pull/40)
```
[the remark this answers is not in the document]
zhiyuanliang-ms: Removed.

releaseNotes/JavaScriptFeatureManagement.md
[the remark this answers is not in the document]
zhiyuanliang-ms: I want to deliver a message that our stable release now can works both in node.js and non-node.js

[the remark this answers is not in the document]
zhenlan: Yes, we didn't say this is node only package. Users should assume both.

zhenlan (approved):

PR #974, opened 2024-09-26
Adds release notes for v4.0.0

rossgrambo:


releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -2,6 +2,50 @@
...
+
+#### Variants
+
+```csharp
+Variant variant = await featureManager.GetVariantAsync(MyFeatureFlags.HelpText);
+
+model.Text = variant.Configuration.Value;
+```
+
+Variants are a tool that can be used to surface different variations of a feature to different segments of an audience. Previously, this library only worked with flags. The flags were limited to boolean values, as they are either enabled or disabled. Variants have dynamic values. They can be any valid json value- like a string, int, or complex object.
```
zhenlan, line 17: This sounded like variants are not feature flags anymore. Variants are an enhancement to feature flags. Feature flags with variants are still flags.

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -2,6 +2,50 @@
...
+        "enabled": true
+    }
+}
+```
+
+Telemetry provides observability into flag evaluations, offering insights into which users received specific flag results. This enables more powerful metric analysis, such as experimentation.
+
+For more details on Telemetry, see [here](https://learn.microsoft.com/en-us/azure/azure-app-configuration/feature-management-dotnet-reference#telemetry).
+
+### Migration
```
zhenlan, line 37: There are no breaking changes, so there are no migration instructions necessary. I don't think we need this section.

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -2,6 +2,50 @@
...
+
+**No changes are required** When migrating from 3.5.0 to 4.0.0.
+
+1. In order to use variants or cancellation tokens, change references from `IFeatureManager` to `IVariantFeatureManager`.
+
+### Additional Changes
+
+* Added `IVariantFeatureManager` which is the successor of the existing `IFeatureManager`. It continues to offer the functions of `IFeatureManager`, but offers the new `GetVariantAsync` methods and offers `CancellationToken` on all methods.
```
zhenlan, line 45: This should be just part of the release note for the "variant feature flag". The `IVariantFeatureManager` is one of many APIs we introduced. The `CancellationToken` support is how the API should be. If you think it's important to call out that `IVariantFeatureManager` is the successor of the existing `IFeatureManager`, call it out in the documentation, which has way more viewers than this release note.

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -2,6 +2,50 @@
...
+* Added `IVariantServiceProvider` which enables variants to control which service is added in DI. See more details [here](https://learn.microsoft.com/en-us/azure/azure-app-configuration/feature-management-dotnet-reference#variants-in-dependency-injection).
```
zhenlan, line 46: This should be a separate item under enhancement. There are multiple APIs involved for this feature to work, so I would focus on the description of the feature instead of just this one API.

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -2,6 +2,50 @@
...
+* Added support for variant feature flags defined using [Microsoft Feature Management schema](https://github.com/microsoft/FeatureManagement/blob/main/Schema/FeatureManagement.v2.0.0.schema.json). Variants and telemetry can be declared using [Microsoft Feature Flag schema v2](https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json). [Sample](https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/examples/VariantAndTelemetryDemo/appsettings.json#L12).
```
zhenlan, line 47: I would just list this under the enhancement instead of hiding it under some additional changes :)

releaseNotes/Microsoft.Featuremanagement.md
rossgrambo: Actually there is one- I'll update the PR to include it. It's fairly rare- so perhaps it doesn't need its own section.

```
It should be uncommon, but when manually constructing `FeatureTagHelper`, you'll now need to pass both an `IFeatureManager` and an `IVariantFeatureManager`.
```

releaseNotes/Microsoft.Featuremanagement.md
rossgrambo: Hmm. I agree. This was taken from the `preview1` release notes but it does sound a little incorrect. How about:

```
This library now supports feature flags with variants. Variants are a tool that can be used to surface different variations of a feature to different segments of an audience. Previously, flags were limited to boolean values, as they are either enabled or disabled. Variants have dynamic values. There can be more than two and they can be any valid json value- like a string, int, or complex object.
```

releaseNotes/Microsoft.Featuremanagement.md
rossgrambo: It feels separate from the Variants section because of the cancellation / successor part. So I promoted it to it's own section.

rossgrambo:
How do we want to handle release notes on `Microsoft.Featuremanagement.Telemetry.ApplicationInsights.AspNetCore` (The deprecated package) ?

zhenlan:
> How do we want to handle release notes on `Microsoft.Featuremanagement.Telemetry.ApplicationInsights.AspNetCore` (The deprecated package) ?

How about updating its release notes one more time with a note saying this package has been deprecated and its functionalities have been moved to the xxxxx package?

releaseNotes/Microsoft.Featuremanagement.md
zhenlan: If there is a breaking change, small or big, we should then have a breaking change section with a small explanation and migration instruction.

I don't follow the change you said. I guess you can use me to test if your instruction is clear once you have it :)

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -2,6 +2,64 @@
...
+For more details on Variants, see [here](https://learn.microsoft.com/en-us/azure/azure-app-configuration/feature-management-dotnet-reference#variants).
+
+#### Variant Service Provider
+
+```csharp
+IVariantServiceProvider<IAlgorithm> algorithmServiceProvider;
+...
+
+IAlgorithm forecastAlgorithm = await algorithmServiceProvider.GetServiceAsync(cancellationToken);
+```
```
zhenlan, line 32: Honestly speaking, without reading the document, these two lines don't mean much. I would just drop it and let people read the document.

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -2,6 +2,64 @@
...
+
+Variant feature flags can be used in conjunction with dependency injection to surface different implementations of a service for different users.
```
zhenlan, line 34: ... This is accomplished by using the Variant Service Provider.

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -2,6 +2,64 @@
...
+
+Telemetry provides observability into flag evaluations, offering insights into which users received specific flag results. This enables more powerful metric analysis, such as experimentation.
+
+For more details on Telemetry, see [here](https://learn.microsoft.com/en-us/azure/azure-app-configuration/feature-management-dotnet-reference#telemetry).
+
+#### Microsoft Feature Management Schema
+
+Added support for variant feature flags defined using [Microsoft Feature Management schema](https://github.com/microsoft/FeatureManagement/blob/main/Schema/FeatureManagement.v2.0.0.schema.json). Variants and telemetry can be declared using [Microsoft Feature Flag schema v2](https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json). [Sample](https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/examples/VariantAndTelemetryDemo/appsettings.json#L12).
+
+### Migration
```
zhenlan, line 58: If it's breaking change, let's call it breaking change.

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -2,6 +2,64 @@
...
+
+1. It should be uncommon, but when manually constructing `FeatureTagHelper`, you'll now need to pass both an `IFeatureManager` and an `IVariantFeatureManager`.
+1. In order to use variants or cancellation tokens, change references from `IFeatureManager` to `IVariantFeatureManager`.
```
zhenlan, line 61: I would remove it. We've talked about cancellation token already.

rossgrambo:
Resolved all comments

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -2,6 +2,45 @@
...
+
+For more details on Telemetry, see [here](https://learn.microsoft.com/en-us/azure/azure-app-configuration/feature-management-dotnet-reference#telemetry).
+
+#### Microsoft Feature Management Schema
+
+Added support for variant feature flags defined using [Microsoft Feature Management schema](https://github.com/microsoft/FeatureManagement/blob/main/Schema/FeatureManagement.v2.0.0.schema.json). Variants and telemetry can be declared using [Microsoft Feature Flag schema v2](https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json). Here is a [Sample](https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/examples/VariantAndTelemetryDemo/appsettings.json#L12).
+
+### Breaking Changes
+
+1. The constructor of `FeatureTagHelper` has been adjusted to take both an `IFeatureManager` and an `IVariantFeatureManager` for backwards compatability. Most use cases don't directly use the constructor, but if the constructor is directly called, the call will need to be adjusted to pass the additional parameter.
```
zhenlan, line 42: - There is only one item. It's odd to have an index of 1.
- We are saying we added a breaking change "for backwards compatibility". It doesn't make sense.
- I still don't grasp the breaking change. Can we add before and after?

releaseNotes/Microsoft.Featuremanagement.Telemetry.ApplicationInsights.md
```diff
@@ -2,6 +2,21 @@
...
+### Enhancements
+
+#### Application Insights
+
+```csharp
+builder.Services.AddFeatureManagement()
+    .AddApplicationInsightsTelemetry();
+```
+
+To export telemetry to Application Insights as custom events, we offer a new builder method `IFeatureManagementBuilder.AddApplicationInsightsTelemetry()`. This causes outgoing Application Insights telemetry to be tagged with TargetingId and will emit evaluation events to the custom events table- using the `TelemetryClient` if available in DI.
```
zhenlan, line 16: Too much implementation detail. Users should read documents for that level of detail.

#### Application Insights

Added support for Application Insights telemetry. To publish feature flag evaluation data to Application Insights, register the Application Insights telemetry publisher as shown below.

```csharp
builder.Services.AddFeatureManagement()
       .AddApplicationInsightsTelemetry();
```

releaseNotes/Microsoft.Featuremanagement.Telemetry.ApplicationInsights.AspNetCore.md
```diff
@@ -2,6 +2,16 @@
 
 [Source code][source_code] | [Package (NuGet)][package] | [Samples][samples] | [Product documentation][docs]
 
+## 4.0.0 - September 24, 2024
```
zhenlan, line 5: We don't have a release a 4.0.0 to deprecate it, right? So just have the following at the beginning of this release note.

## Deprecation

This package has been deprecated and it's functionality is now available as a part of the [Microsoft.FeatureManagement.ApplicationInsights](https://github.com/Azure/AppConfiguration/blob/main/releaseNotes/Microsoft.Featuremanagement.Telemetry.ApplicationInsights.md) package.

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -2,6 +2,45 @@
 
 [Source code][source_code] | [Package (NuGet)][package] | [Samples][samples] | [Product documentation][docs]
 
+## 4.0.0 - September 24, 2024
+
+### Enhancements
+
+#### Variant Feature Flags
```
zhenlan, line 9: BTW, I think it will be helpful to call out that the variant feature flags require v8 of the provider library to work if users load feature flags from App Configuration.

releaseNotes/Microsoft.Featuremanagement.md
rossgrambo:
```csharp
IFeatureManager featureManager = serviceProvider.GetRequiredService<IFeatureManager>();
FeatureTagHelper a = new FeatureTagHelper(featureManager);
```

Has become
```csharp
FeatureManager featureManager = serviceProvider.GetRequiredService<FeatureManager>();
FeatureTagHelper a = new FeatureTagHelper(featureManager, featureManager); 
// Or the more explicit
IFeatureManager featureManager = serviceProvider.GetRequiredService<IFeatureManager>();
IVariantFeatureManager variantFeatureManager = serviceProvider.GetRequiredService<IVariantFeatureManager>();
FeatureTagHelper a = new FeatureTagHelper(featureManager, variantFeatureManager);
```

This makes it backwards compatible with a custom `IFeatureManager`. While the variant logic it now has uses the `IVariantFeatureManager`.

releaseNotes/Microsoft.Featuremanagement.Telemetry.ApplicationInsights.md
rossgrambo: We can simplify it if you prefer. But I think we need to call out that all outgoing events will have extra data appended to them. It's not only the feature evaluation data. (aka- if someone uses this but never calls `IsEnabledAsync()/GetVariantAsync()`- there's still a change)

zhenlan: The TargetingId is not even added by the AddApplicationInsightsTelemetry API, is it? Let users read documentation.

rossgrambo: It is. Calling `.AddApplicationInsightsTelemetry()` adds the `TargetingTelemetryInitializer`. This appends `TargetingId` onto all outgoing events as long as `TargetingId` is available from the `ITargetingContextAccessor`.

How about we just adjust your version to say:
> Added support for Application Insights telemetry. To publish feature flag evaluation data and tag outgoing events with targeting information, register the Application Insights telemetry publisher as shown below.

releaseNotes/Microsoft.Featuremanagement.md
rossgrambo: > There is only one item. It's odd to have an index of 1.

I prefer the index because it makes it clear that there is only a single breaking change. A paragraph sounds like there might be multiple. Could change the header to be singular but I prefer consistency. 

> We are saying we added a breaking change "for backwards compatibility". It doesn't make sense.

Unfortunately it's what's happening. We can leave it out if it makes it hard to read/understand.

releaseNotes/Microsoft.Featuremanagement.Telemetry.ApplicationInsights.AspNetCore.md
rossgrambo: Agreed

releaseNotes/Microsoft.Featuremanagement.md
zhenlan: It may give you the opposite of what you intend. With an index of 1, it confuses your readers and they think you forgot to write 2 and more. You can use a bullet point instead.

The `FeatureTagHelper` belongs to the MS.FM.AspNetCore package, so this is the wrong place to add this note.

- What's the scenario one will call the constructor directly?
- The feature tag can now be controlled by variant feature flags. I will certainly list this under enhancement. We are not making this change for the backward compatibility. Didn't we have a document for this?

rossgrambo: > What's the scenario one will call the constructor directly?

Searching github shows a few results of people manually creating it for tests. I'm unaware of any other reason to call it directly. 

> The feature tag can now be controlled by variant feature flags. I will certainly list this under enhancement.

It was already released in preview4 and was a breaking change for customers using `IFeatureManager`- because it was adjusted to take `IVariantFeatureManager` instead. In this release- we're adjusting it to take both. This means customers using custom IFeatureManagers are still able to use their custom interface. However, it continues to be a breaking change from v3 since the constructor is changing.

I've been trying to call out the breaking change from `preview4` to `stable`. We added the extra parameter to support what we had in preview3 and before. (backwards compatibility that was lost on preview4). I can instead call out the breaking change from v3 -> v4 since this is kind of a recap of the preview changes.

How about:

MS.FM.AspNetCore

### Enhancements

#### FeatureTagHelper variants

FeatureTagHelper now supports variants. This allows MVC views to use the `<feature>` tag to conditionally render content based on whether a specific variant of a feature is assigned.

``` HTML+Razor
<feature name="FeatureX" variant="Alpha">
  <p>This can only be seen if variant 'Alpha' of 'FeatureX' is assigned.</p>
</feature>
```

For more details on ASP.NET views and variants, see [here](https://learn.microsoft.com/en-us/azure/azure-app-configuration/feature-management-dotnet-reference#view).

### Breaking Change

* The constructor for `FeatureTagHelper` now requires both an `IFeatureManager` and an `IVariantFeatureManager`. While this class is typically not instantiated directly, any direct instantiation will need to be updated.

releaseNotes/Microsoft.Featuremanagement.Telemetry.ApplicationInsights.md
```diff
@@ -2,6 +2,21 @@
...
+#### Application Insights
+
+Added support for Application Insights telemetry. To publish feature flag evaluation data and tag outgoing events with targeting information, register the Application Insights telemetry publisher as shown below.
+
+```csharp
+builder.Services.AddFeatureManagement()
+    .AddApplicationInsightsTelemetry();
+```
+
+For more details on Application Insights Publishing, see [here](https://learn.microsoft.com/en-us/azure/azure-app-configuration/feature-management-dotnet-reference#application-insights-telemetry-publisher)
```
jimmyca15, line 18: Does this link work for you consistently. It seemed to work for me once then break when I viewed it again as if it depended upon my previous tab state on the page.

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -2,6 +2,47 @@
...
+
+#### Telemetry
+
+Telemetry provides observability into flag evaluations, offering insights into which users received specific flag results. This enables more powerful metric analysis, such as experimentation.
+
+For more details on Telemetry, see [here](https://learn.microsoft.com/en-us/azure/azure-app-configuration/feature-management-dotnet-reference#telemetry).
+
+#### Microsoft Feature Management Schema
+
+Added support for variant feature flags defined using [Microsoft Feature Management schema](https://github.com/microsoft/FeatureManagement/blob/main/Schema/FeatureManagement.v2.0.0.schema.json). Variants and telemetry can be declared using [Microsoft Feature Flag schema v2](https://github.com/Azure/AppConfiguration/blob/main/docs/FeatureManagement/FeatureFlag.v2.0.0.schema.json). Here is a [Sample](https://github.com/microsoft/FeatureManagement-Dotnet/blob/main/examples/VariantAndTelemetryDemo/appsettings.json#L12).
```
jimmyca15, line 40: The Microsoft Feature Flag schema v2 link is broken for me.

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -2,6 +2,47 @@
...
+
+### Breaking Changes
+
+1. The constructor of `FeatureTagHelper` has been adjusted to take both an `IFeatureManager` and an `IVariantFeatureManager` for backwards compatability. Most use cases don't directly use the constructor, but if the constructor is directly called, the call will need to be adjusted to pass the additional parameter.
```
jimmyca15, line 44: Breaking change for backward compatibility is a funny thing to read. I would just list this as a bullet point since the only reason we took the breaking change is we expect it to be inconsequential.

Something like

* The `FeatureTagHelper` constructor now requires an `IVariantFeatureManager`.

releaseNotes/Microsoft.Featuremanagement.md
zhenlan: Most customers won't try the preview releases or track the breaking changes in preview releases. Therefore, for every stable release, we should include (repeat) all enhancements, bug fixes and breaking changes. We shouldn't force customers to read all the notes of preview releases. There is no way to tell whether something makes it to the final stable release or not if we don't call them out explicitly.

Please include everything since the last stable release if there is anything else missing.

releaseNotes/Microsoft.Featuremanagement.Telemetry.ApplicationInsights.md
```diff
@@ -2,6 +2,21 @@
 
 [Source code][source_code] | [Package (NuGet)][package] | [Samples][samples] | [Product documentation][docs]
 
+## 4.0.0 - September 24, 2024
+
+### Enhancements
+
+#### Application Insights
```
zhenlan, line 9: This whole package is named after ApplicationInsights. I don't see a value in having this subsection. We could just put your content below in bullet point.

releaseNotes/Microsoft.Featuremanagement.md
rossgrambo: Our release notes guide calls out that we should [explain why a breaking change is made and how to migrate](https://microsoft.sharepoint.com/teams/azconfig/_layouts/OneNote.aspx?id=%2Fteams%2Fazconfig%2FSiteAssets%2FAzure%20Configuration%20Service%20Notebook&wd=target%28HowTos.one%7C0F93BC29-5F6A-4F4A-8215-ABDAD82CE46B%2FWrite%20release%20notes%7CBBF2648C-3EBF-406A-B721-EF694BC5F733%2F%29onenote:https://microsoft.sharepoint.com/teams/azconfig/SiteAssets/Azure%20Configuration%20Service%20Notebook/HowTos.one#Write%20release%20notes&section-id={0F93BC29-5F6A-4F4A-8215-ABDAD82CE46B}&page-id={BBF2648C-3EBF-406A-B721-EF694BC5F733}&object-id={982A1903-F7CC-0749-27C0-2D7D4C467173}&4F).

How about:
* The `FeatureTagHelper` constructor now requires an `IVariantFeatureManager` to support new variant functionality. While this class is typically not instantiated directly, any direct instantiation will need to be updated.

jimmyca15: I'm good with that summary.

releaseNotes/Microsoft.Featuremanagement.md
rossgrambo: Updated both links to working permalinks

releaseNotes/Microsoft.Featuremanagement.Telemetry.ApplicationInsights.md
rossgrambo: Interesting, I thought it was just me having that problem... It always a problem even if you include the `?pivots=preview-version`. I suspect the problem will go away once the pivots are gone.

releaseNotes/Microsoft.Featuremanagement.AspNetCore.md
```diff
@@ -2,6 +2,26 @@
...
+* Added support for variants in `FeatureTagHelper`. This allows MVC views to use the `<feature>` tag to conditionally render content based on whether a specific variant of a feature is assigned.
+
+    ``` HTML+Razor
+    <feature name="FeatureX" variant="Alpha">
+      <p>This can only be seen if variant 'Alpha' of 'FeatureX' is assigned.</p>
+    </feature>
+    ```
+
+    For more details on ASP.NET views and variants, see [here](https://learn.microsoft.com/en-us/azure/azure-app-configuration/feature-management-dotnet-reference#view).
+
```
zhenlan, line 20: We should still keep the note, right?

Updated `Microsoft.FeatureManagement` reference to `4.0.0`...

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -2,6 +2,47 @@
...
+The new `IVariantFeatureManager` has been introduced as the successor to the existing `IFeatureManager`. It retains all the functionalities of `IFeatureManager` while adding new `GetVariantAsync` methods and supporting `CancellationToken` for all methods.
+
+``` C#
+IVariantFeatureManager featureManager;
+...
+Variant variant = await featureManager.GetVariantAsync(MyFeatureFlags.HelpText, CancellationToken.None);
+model.Text = variant.Configuration.Value;
+```
+
+*Note: If reading variant flags from App Configuration, version `8.0.0` or above for the `Microsoft.Azure.AppConfiguration` or `Microsoft.Azure.AppConfiguration.AspNetCore` package is required.*
```
zhenlan, line 22: > `Microsoft.Azure.AppConfiguration`

There is no such package :)

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -2,6 +2,47 @@
...
+
+Telemetry provides observability into flag evaluations, offering insights into which users received specific flag results. This enables more powerful metric analysis, such as experimentation.
+
+For more details on Telemetry, see [here](https://learn.microsoft.com/en-us/azure/azure-app-configuration/feature-management-dotnet-reference#telemetry).
+
+#### Microsoft Feature Management Schema
+
+Added support for variant feature flags defined using [Microsoft Feature Management schema](https://github.com/microsoft/FeatureManagement/blob/c5fab16dbf1450dce0bbfe7c4207da735ff31916/Schema/FeatureManagement.v2.0.0.schema.json). Variants and telemetry can be declared using [Microsoft Feature Flag schema v2](https://github.com/microsoft/FeatureManagement/blob/c5fab16dbf1450dce0bbfe7c4207da735ff31916/Schema/FeatureFlag.v2.0.0.schema.json). Here is a [Sample](https://github.com/microsoft/FeatureManagement-Dotnet/blob/f47e188babea0a91488d2e6a0b2ab4c9405d0794/examples/VariantAndTelemetryDemo/appsettings.json#L12).
+
+### Breaking Changes
```
zhenlan, line 42: Remove

zhenlan (approved):

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -39,6 +39,12 @@ For more details on Telemetry, see [here](https://learn.microsoft.com/en-us/azur
 
 Added support for variant feature flags defined using [Microsoft Feature Management schema](https://github.com/microsoft/FeatureManagement/blob/c5fab16dbf1450dce0bbfe7c4207da735ff31916/Schema/FeatureManagement.v2.0.0.schema.json). Variants and telemetry can be declared using [Microsoft Feature Flag schema v2](https://github.com/microsoft/FeatureManagement/blob/c5fab16dbf1450dce0bbfe7c4207da735ff31916/Schema/FeatureFlag.v2.0.0.schema.json). Here is a [Sample](https://github.com/microsoft/FeatureManagement-Dotnet/blob/f47e188babea0a91488d2e6a0b2ab4c9405d0794/examples/VariantAndTelemetryDemo/appsettings.json#L12).
 
+#### Additional Changes
+
+* Optimized memory allocation for calls to `IsEnabledAsync` and `GetVariantAsync`.
```
zhenlan, line 44: I thought it's not just memory. Do we have numbers in terms of much improvement?

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -39,6 +39,12 @@ For more details on Telemetry, see [here](https://learn.microsoft.com/en-us/azur
...
+* Added support for .NET 9
+* Removed support for .NET 7
```
zhenlan, line 46: This should be categorized under breaking change.

zhenlan: And I don't think this belongs to the MS.FM library.

releaseNotes/Microsoft.Featuremanagement.md
rossgrambo: The goal was to reduce memory allocation. Memory is now always more performant than v3.

With speed it's not as straightforward. v4 in .net6 is still slower than v3, while v4 in .net8 is faster than v3. I chose to omit it since it wasn't as easily stated.

releaseNotes/Microsoft.Featuremanagement.md
rossgrambo: For MS.FM- .net 7 was removed from tests. Meaning we are no longer ensuring there's not a .net 7 specific breaking change. 

For MS.FM.ASPNET- .net 7 was removed as a target from the actual library and from the tests. This means we no longer run tests against it and there's won't be a .net 7 specific build within the nuget package when installing MS.FM.ASPNET.  

There's no breaking change since the .net 6 build will continue to work in .net 7 environments.

rossgrambo: I'll add

* Removed support for .NET 7

In the ASP.NET library as well.

releaseNotes/Microsoft.Featuremanagement.md
zhenlan: .NET 6 is deprecated soon. What's the perf improvement in .net8?

releaseNotes/Microsoft.Featuremanagement.AspNetCore.md
```diff
@@ -2,6 +2,28 @@
...
+    </feature>
+    ```
+
+    For more details on ASP.NET views and variants, see [here](https://learn.microsoft.com/en-us/azure/azure-app-configuration/feature-management-dotnet-reference#view).
+* Updated `Microsoft.FeatureManagement` reference to `4.0.0`. See the [release notes](./Microsoft.Featuremanagement.md) for more information on the changes.
+
+### Breaking Change
+
+* The `FeatureTagHelper` constructor now requires an `IVariantFeatureManager` to support new variant functionality. While this class is typically not instantiated directly, any direct instantiation will need to be updated.
+* Removed direct support for .NET 7. The .NET 6 library will continue to work in .NET 7 applications.
```
zhenlan, line 25: > The .NET 6 library will continue to work in .NET 7 applications.

Yes, theoretically, but we don't need to endorse it. Suggest below:

Removed .NET 7 as a target framework as .NET 7 is out of support.

releaseNotes/Microsoft.Featuremanagement.md
```diff
@@ -2,6 +2,49 @@
...
+For more details on Telemetry, see [here](https://learn.microsoft.com/en-us/azure/azure-app-configuration/feature-management-dotnet-reference#telemetry).
+
+#### Microsoft Feature Management Schema
+
+Added support for variant feature flags defined using [Microsoft Feature Management schema](https://github.com/microsoft/FeatureManagement/blob/c5fab16dbf1450dce0bbfe7c4207da735ff31916/Schema/FeatureManagement.v2.0.0.schema.json). Variants and telemetry can be declared using [Microsoft Feature Flag schema v2](https://github.com/microsoft/FeatureManagement/blob/c5fab16dbf1450dce0bbfe7c4207da735ff31916/Schema/FeatureFlag.v2.0.0.schema.json). Here is a [Sample](https://github.com/microsoft/FeatureManagement-Dotnet/blob/f47e188babea0a91488d2e6a0b2ab4c9405d0794/examples/VariantAndTelemetryDemo/appsettings.json#L12).
+
+#### Additional Changes
+
+* Optimized memory allocation for calls to `IsEnabledAsync` and `GetVariantAsync`.
+* Added support for .NET 9
```
zhenlan, line 45: Same as .NET 7. This one shouldn't live here.

releaseNotes/Microsoft.Featuremanagement.md
zhenlan: This is releasing note of prod. Testing is irrelevant. Please remove.

releaseNotes/Microsoft.Featuremanagement.AspNetCore.md
```diff
@@ -2,6 +2,28 @@
...
+
+    ``` HTML+Razor
+    <feature name="FeatureX" variant="Alpha">
+      <p>This can only be seen if variant 'Alpha' of 'FeatureX' is assigned.</p>
+    </feature>
+    ```
+
+    For more details on ASP.NET views and variants, see [here](https://learn.microsoft.com/en-us/azure/azure-app-configuration/feature-management-dotnet-reference#view).
+* Updated `Microsoft.FeatureManagement` reference to `4.0.0`. See the [release notes](./Microsoft.Featuremanagement.md) for more information on the changes.
+
```
zhenlan, line 21: I thought you wanted to add this?

Added support for .NET 9 as a target framework.

zhenlan (approved):

PR #1026, opened 2025-03-04
Add Function App example with .NET 8 in-process model

zhenlan:
The in-process model will be supported until November 10, 2026.


examples/DotNetCore/AzureFunctions/FunctionAppInProcess/ShowBetaFeature.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        [FunctionName("ShowBetaFeature")]
+        public async Task<IActionResult> Run(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post", Route = null)] HttpRequest req,
+            ILogger log)
+        {
+            log.LogInformation("C# HTTP trigger function processed a request.");
+
+            // Signal to refresh the feature flags from Azure App Configuration.
+            // This will be a no-op if the refresh interval time window is not reached.
```
avanigupta, line 26:
```suggestion
            // This will be a no-op if the refresh interval has not elapsed.

```

avanigupta (approved):

PR #1055, opened 2025-05-20
Add a .NET ChatApp example for AI configuration

zhenlan:
Example output of the ChatApp:

image

### DIFF
--- a/examples/DotNetCore/ChatApp/ChatApp/Message.cs
+++ b/examples/DotNetCore/ChatApp/ChatApp/Message.cs
@@ -1,13 +1,16 @@
-﻿using Microsoft.Extensions.Configuration;
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using Microsoft.Extensions.Configuration;
 
 namespace ChatApp
 {
     internal class Message
     {
         [ConfigurationKeyName("role")]
-        public string Role { get; set; }
+        public required string Role { get; set; }
 
         [ConfigurationKeyName("content")]
-        public string Content { get; set; }
+        public string? Content { get; set; }
     }
 }
--- a/examples/DotNetCore/ChatApp/ChatApp/Message.cs
+++ b/examples/DotNetCore/ChatApp/ChatApp/Message.cs
@@ -1,0 +1,13 @@
+﻿using Microsoft.Extensions.Configuration;
+
+namespace ChatApp
+{
+    internal class Message
+    {
+        [ConfigurationKeyName("role")]
+        public string Role { get; set; }
+
+        [ConfigurationKeyName("content")]
+        public string Content { get; set; }
+    }
+}
--- a/examples/DotNetCore/ChatApp/ChatApp/ModelConfiguration.cs
+++ b/examples/DotNetCore/ChatApp/ChatApp/ModelConfiguration.cs
@@ -1,0 +1,32 @@
+﻿using Microsoft.Extensions.Configuration;
+using OpenAI.Chat;
+
+namespace ChatApp
+{
+    internal class ModelConfiguration
+    {
+        [ConfigurationKeyName("model")]
+        public string Model { get; set; }
+
+        [ConfigurationKeyName("messages")]
+        public List<Message> Messages { get; set; }
+
+        [ConfigurationKeyName("max_tokens")]
+        public int MaxTokens { get; set; }
+
+        [ConfigurationKeyName("temperature")]
+        public float Temperature { get; set; }
+
+        [ConfigurationKeyName("top_p")]
+        public float TopP { get; set; }
+
+        public IEnumerable<ChatMessage> ChatMessages =>
+            Messages.Select<Message, ChatMessage>(message => message.Role switch
+            {
+                "system" => ChatMessage.CreateSystemMessage(message.Content),
+                "user" => ChatMessage.CreateUserMessage(message.Content),
+                "assistant" => ChatMessage.CreateAssistantMessage(message.Content),
+                _ => throw new ArgumentException($"Unknown role: {message.Role}", nameof(message.Role))
+            });
+    }
+}
--- a/examples/DotNetCore/ChatApp/ChatApp/ModelConfiguration.cs
+++ b/examples/DotNetCore/ChatApp/ChatApp/ModelConfiguration.cs
@@ -1,15 +1,17 @@
-﻿using Microsoft.Extensions.Configuration;
-using OpenAI.Chat;
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using Microsoft.Extensions.Configuration;
 
 namespace ChatApp
 {
     internal class ModelConfiguration
     {
         [ConfigurationKeyName("model")]
-        public string Model { get; set; }
+        public string? Model { get; set; }
 
         [ConfigurationKeyName("messages")]
-        public List<Message> Messages { get; set; }
+        public List<Message>? Messages { get; set; }
 
         [ConfigurationKeyName("max_tokens")]
         public int MaxTokens { get; set; }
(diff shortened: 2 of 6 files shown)


examples/DotNetCore/ChatApp/ChatApp/ModelConfiguration.cs
```diff
@@ -0,0 +1,32 @@
...
+        [ConfigurationKeyName("max_tokens")]
+        public int MaxTokens { get; set; }
+
+        [ConfigurationKeyName("temperature")]
+        public float Temperature { get; set; }
+
+        [ConfigurationKeyName("top_p")]
+        public float TopP { get; set; }
+
+        public IEnumerable<ChatMessage> ChatMessages =>
```
jimmyca15, line 23: I would keep the configuration type simple POCO.

examples/DotNetCore/ChatApp/ChatApp/Program.cs
```diff
@@ -0,0 +1,66 @@
...
+IConfiguration configuration = new ConfigurationBuilder()
+    .AddAzureAppConfiguration(options =>
+    {
+        Uri endpoint = new(Environment.GetEnvironmentVariable("AZURE_APPCONFIG_ENDPOINT") ??
+            throw new InvalidOperationException("The environment variable 'AZURE_APPCONFIG_ENDPOINT' is not set or is empty."));
+        options.Connect(endpoint, credential)
+               // Load all keys that have a specific prefix and no label.
+               .Select("ChatLLM:*")
+               // Reload configuration if any selected key-values have changed.
+               // Use the default refresh interval of 30 seconds. It can be overridden via refreshOptions.SetRefreshInterval.
```
jimmyca15, line 22: This comment is outdated.

examples/DotNetCore/ChatApp/ChatApp/Program.cs
```diff
@@ -0,0 +1,66 @@
...
+IConfigurationRefresher _refresher = null;
+
+// Load configuration from Azure App Configuration
+IConfiguration configuration = new ConfigurationBuilder()
+    .AddAzureAppConfiguration(options =>
+    {
+        Uri endpoint = new(Environment.GetEnvironmentVariable("AZURE_APPCONFIG_ENDPOINT") ??
+            throw new InvalidOperationException("The environment variable 'AZURE_APPCONFIG_ENDPOINT' is not set or is empty."));
+        options.Connect(endpoint, credential)
+               // Load all keys that have a specific prefix and no label.
```
jimmyca15, line 19: If we want to keep this comment, I think it could be more specific. Loads all keys that start with "ChatLLM:"

jimmyca15: I also might consider to change the configuration a bit so that the prefix is more of an application name. `ChatApp:`

`ChatApp:AzureOpenAi:Endpoint`
`ChatApp:AzureOpenAi:DeploymentName`

`ChatApp:Model` or `ChatApp:LLM`

examples/DotNetCore/ChatApp/ChatApp/Program.cs
```diff
@@ -0,0 +1,66 @@
...
+AzureOpenAIClient azureClient = new(chatEndpoint, credential);
+ChatClient chatClient = azureClient.GetChatClient(deploymentName);
+
+while (true)
+{
+    // Refresh the configuration from Azure App Configuration
+    await _refresher.TryRefreshAsync();
+
+    // Configure chat completion with AI configuration
+    var modelConfig = configuration.GetSection("ChatLLM:Model").Get<ModelConfiguration>();
```
jimmyca15, line 47: for a sample I'd rather go with full `configuration`
```suggestion
    var modelConfiguration = configuration.GetSection("ChatLLM:Model").Get<ModelConfiguration>();
```

examples/DotNetCore/ChatApp/ChatApp/Program.cs
```diff
@@ -0,0 +1,66 @@
...
+        TopP = modelConfig.TopP
+    };
+
+    foreach (var message in modelConfig.Messages)
+    {
+        Console.WriteLine($"{message.Role}: {message.Content}");
+    }
+
+    // Get chat response from AI
+    var response = chatClient.CompleteChat(modelConfig.ChatMessages, requestOptions);
```
jimmyca15, line 61: Does this have an async variant?

examples/DotNetCore/ChatApp/ChatApp/Program.cs
```diff
@@ -0,0 +1,66 @@
...
+    await _refresher.TryRefreshAsync();
+
+    // Configure chat completion with AI configuration
+    var modelConfig = configuration.GetSection("ChatLLM:Model").Get<ModelConfiguration>();
+    var requestOptions = new ChatCompletionOptions()
+    {
+        MaxOutputTokenCount = modelConfig.MaxTokens,
+        Temperature = modelConfig.Temperature,
+        TopP = modelConfig.TopP
+    };
```
jimmyca15, line 53: Reflecting on this -- This will be another critical part to help users set up. We're generating the type, and the SDK property names they map to may not be 1:1

examples/DotNetCore/ChatApp/ChatApp/Program.cs
```diff
@@ -0,0 +1,66 @@
...
+    foreach (var message in modelConfig.Messages)
+    {
+        Console.WriteLine($"{message.Role}: {message.Content}");
+    }
+
+    // Get chat response from AI
+    var response = chatClient.CompleteChat(modelConfig.ChatMessages, requestOptions);
+    System.Console.WriteLine($"AI response: {response.Value.Content[0].Text}");
+
+    Console.WriteLine("Press Enter to continue...");
```
jimmyca15, line 64: If interested, we could have this auto-run on configuration update rather than requiring input from the user.

examples/DotNetCore/ChatApp/ChatApp/Program.cs
zhenlan: Oops, I forgot to delete the refresh overriding code. I meant to use the default. Thanks for catching it.

examples/DotNetCore/ChatApp/ChatApp/Program.cs
zhenlan: Yes, let me switch to the async version.

examples/DotNetCore/ChatApp/ChatApp/Program.cs
zhenlan: We need to see how we can keep up the changes.

The max_tokens is deprecated and replaced with max_completion_tokens, but the SDK named the property as MaxOutputTokenCount :(

BTW, the "function" role is also being deprecated and replaced with "tools".

We need to update our Copilot instructions to keep up with these changes and update all templates/UI/code periodically.

zhenlan:
> So this one went with the approach where the user prompt is part of configuration right?

Yes, the user prompt is also from the configuration. Do you prefer it input at runtime?

examples/DotNetCore/ChatApp/ChatApp/Message.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using Microsoft.Extensions.Configuration;
+
+namespace ChatApp
+{
+    internal class Message
+    {
+        [ConfigurationKeyName("role")]
```
MaryanneNjeri, line 10: Since the property name is "Role" and the JSON property name is also "role", we may not need to have the `ConfigurationKeyName`, do we have it for consistency ?

jimmyca15:
> > So this one went with the approach where the user prompt is part of configuration right?
> 
> Yes, the user prompt is also from the configuration. Do you prefer it input at runtime?

No, the way you've done it is actually the way I prefer it.

examples/DotNetCore/ChatApp/ChatApp/Message.cs
jimmyca15: I would keep it for consistency.

jimmyca15 (approved):

zhenlan:
@Muksvso, FYI

PR #1056, opened 2025-05-20
Add multi-tenant guidance for .NET

amerjusupovic:


examples/DotNetCore/MultiTenantApplicationSetup/README.md
```diff
@@ -0,0 +1,96 @@
...
+    #### Tenant 1
+
+    | Key | Value |
+    |---- |-------|
+    | Contoso:Name | Contoso Corp. |
+    | Contoso:Color | Azure |
+
+    #### Tenant 2
+
+    | Key | Value |
```
jimmyca15, line 16: Another name we can use similar to Contoso is Fabrikam. I'd recommend that instead of `ExampleTenant` which isn't analogous to contoso.

examples/DotNetCore/MultiTenantApplicationSetup/README.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+4. Use in a DI Service or Controller via `IOptionsSnapshot`.
+
+```cs
+public class MyController : Controller
+{
+        private readonly IOptionsSnapshot<TenantSettings> _tenantSettings;
+
+        public MyController(IOptionsSnapshot<TenantSettings> tenantSettings)
```
jimmyca15, line 83: I'm thinking, is IOptionsSnapshot actually needed? I don't think so because if I remember correctly IConfigureOptions works on plain IOptions.

jimmyca15 (approved):

PR #1057, opened 2025-05-21
Release note for JS FM 2.1.0

zhiyuanliang-ms:


releaseNotes/JavaScriptFeatureManagement.md
```diff
@@ -2,6 +2,8 @@
 
 [Source code][source_code] | [Package (npm)][package] | [Samples][samples]
 
+## 2.1.0 - May 20, 2025
+
```
zhenlan, line 6: We have nothing to say?

zhiyuanliang-ms: Oh. I copied paste the content from other place but it seems like I forget to press ctrl+s, before I push the change.

releaseNotes/JavaScriptFeatureManagement.md
```diff
@@ -2,6 +2,14 @@
...
+### Enhancements
+
+* Stablized the targeting context accessor. You can configure an `ITargetingContextAccessor` when initializing the `FeatureManager` through `FeatureManagerOptions.targetingContextAccessor`. [#93](https://github.com/microsoft/FeatureManagement-JavaScript/pull/93)
+
+* Stablized the feature evaluation event properties. The shape of the object returned by `createFeatureEvaluationEventProperties` contains the following fields:  `"Version"`, `"FeatureName"`, `"Enabled"`, `"TargetingId"`,  `"Variant`", `"VariantAssignmentReason"`, `"DefaultWhenEnabled"` and `"VariantAssignmentPercentage"`.
```
jimmyca15: Is `createFeatureEvaluationEventProperties` a public function?

jimmyca15: I don't remember talking about it, and if it's not public it shouldn't be mentioned.

zhiyuanliang-ms: It is [public](https://github.com/microsoft/FeatureManagement-JavaScript/blob/main/src/feature-management/src/index.ts#L6)

Here is the [implementation](https://github.com/microsoft/FeatureManagement-JavaScript/blob/main/src/feature-management/src/telemetry/featureEvaluationEvent.ts#L16). This function will take the `EvaluationResult` we exposed and generate an object which has the shape defined by the [schema](https://github.com/microsoft/FeatureManagement/blob/main/Schema/FeatureEvaluationEvent/FeatureEvaluationEvent.v1.0.0.schema.json) So this function is kind of useful for those who write their own telemetry publisher, the [`onFeatureEvaluated`](https://github.com/microsoft/FeatureManagement-JavaScript/blob/main/src/feature-management/src/featureManager.ts#L215) callback will take the `EvaluationResult` as input.

I used to implement the same thing in two app insights packages (node and browser), then I decide to expose a function in the base package.

jimmyca15 (approved):

releaseNotes/JavaScriptFeatureManagement.md
jimmyca15:
```suggestion
* Stablized the feature evaluation event properties. These properties can be obtained through the use of `createFeatureEvaluationEventProperties`, and they include:  `"Version"`, `"FeatureName"`, `"Enabled"`, `"TargetingId"`,  `"Variant`", `"VariantAssignmentReason"`, `"DefaultWhenEnabled"` and `"VariantAssignmentPercentage"`.
```

jimmyca15: Arguably the list of included properties is a bit much for release notes. We may have a better place to just link to.

zhiyuanliang-ms: I used to consider put the schema [link](https://github.com/microsoft/FeatureManagement/blob/main/Schema/FeatureEvaluationEvent/FeatureEvaluationEvent.v1.0.0.schema.json) here. But there are some optional fields generated by the provider there. The ideal solution is to add `Version`, `DefaultWhenEnabled`, `VariantAssignmentPercentage` to schema's "required" array. Then, I can update the release note to say we include the required properties of the schema.

PR #1058, opened 2025-05-21
Upgrade the .NET console app sample to .NET 8

zhenlan:
1.	Upgrade from .NET Core 3.1 to .NET 8
2.	Use the minimal project template style
3.	Switch to DefaultAzureCredential for authentication with Azure App Configuration

GitHub Copilot agent did most of the work :)

### DIFF
--- a/examples/DotNetCore/ConsoleApplication/Program.cs
+++ b/examples/DotNetCore/ConsoleApplication/Program.cs
@@ -1,4 +1,7 @@
-﻿using Azure.Identity;
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+//
+using Azure.Identity;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Configuration.AzureAppConfiguration;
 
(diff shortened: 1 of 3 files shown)


examples/DotNetCore/ConsoleApplication/Program.cs
```diff
@@ -1,93 +1,66 @@
...
+    // Use DefaultAzureCredential for Microsoft Entra ID authentication
+    options.Connect(new Uri(endpoint), new DefaultAzureCredential())
+           // Load all keys that start with "Settings:" and have no label.
+           .Select("Settings:*")
+           .TrimKeyPrefix("Settings:")
+           .ConfigureRefresh(refreshOptions =>
+           {
+               // Reload configuration if any selected key-values have changed.
+               refreshOptions.RegisterAll()
+                             .SetRefreshInterval(TimeSpan.FromSeconds(10));
```
jimmyca15, line 37: Since we recently went with the default refresh rate, wondering if we should do it here.

jimmyca15 (approved):

examples/DotNetCore/ConsoleApplication/Program.cs
zhenlan: Makes sense. Updated.

PR #1078, opened 2025-07-14
Add Go chat app sample

linglingye001:


examples/Go/ChatApp/README.md
```diff
@@ -0,0 +1,91 @@
...
+### Example Session
+```
+Chat started! What's on your mind?
+You: Hello, how are you?
+AI: Hello! I'm doing well, thank you for asking. How can I help you today?
+
+You: What can you tell me about machine learning?
+AI: Machine learning is a subset of artificial intelligence that focuses on...
+
+You: exit
```
juniwang: It makes people to think they need to input text `exit`. should use something like `[Press Enter to exit]` or use a screenshot.

linglingye001: Updated, thanks

examples/Go/ChatApp/main.go
```diff
@@ -0,0 +1,224 @@
...
+
+func main() {
+	ctx := context.Background()
+	configProvider, err := loadAzureAppConfiguration(ctx)
+	if err != nil {
+		log.Fatal("Error loading Azure App Configuration:", err)
+		return
+	}
+	
+	// Load AI configuration from Azure App Configuration
```
zhenlan: What you are doing here is to instantiate the AIConfig object from configuration. The data is already loaded from AppConfig in the previous step.

examples/Go/ChatApp/main.go
```diff
@@ -0,0 +1,224 @@
...
+		azure.WithEndpoint(app.aiConfig.AzureOpenAI.Endpoint, app.aiConfig.AzureOpenAI.APIVersion),
+		azure.WithTokenCredential(tokenCredential),
+	)
+
+	app.openAIClient = client
+	return nil
+}
+
+func (app *ChatApp) callAzureOpenAI(userMessage string) (string, error) {
+	messages := []openai.ChatCompletionMessageParamUnion{}
```
zhenlan: The chat history should be stored and sent back to AI for the next round of the conversation. See .NET implementation:
https://github.com/Azure/AppConfiguration/blob/main/examples/DotNetCore/ChatApp/ChatApp/Program.cs#L57

linglingye001: Thanks, updated to store and use chat history.

examples/Go/ChatApp/main.go
```diff
@@ -0,0 +1,232 @@
...
+			// Load all keys that start with "ChatApp:" and have no label
+			{
+				KeyFilter: "ChatApp:*",
+			},
+		},
+		TrimKeyPrefixes: []string{"ChatApp:"},
+		RefreshOptions: azureappconfiguration.KeyValueRefreshOptions{
+			Enabled:  true,
+			Interval: 10 * time.Second,
+		},
```
zhenlan, line 120: This is not an example of the Key Vault reference feature. I don't think auto-refresh is necessary.

(BTW, 10s AKV refresh seems way too aggressive. When we give examples, I would expect something like once an hour or even once a day.)

examples/Go/ChatApp/main.go
```diff
@@ -0,0 +1,232 @@
...
+	APIVersion string
+	APIKey     string
+}
+
+type Message struct {
+	Role    string `json:"role"`
+	Content string `json:"content"`
+}
+
+func loadAzureAppConfiguration(ctx context.Context) (*azureappconfiguration.AzureAppConfiguration, error) {
```
zhenlan, line 97: It will be nice to add a comment to describe what this function does.

examples/Go/ChatApp/main.go
```diff
@@ -0,0 +1,232 @@
...
+func (app *ChatApp) createAzureOpenAIClient() error {
+	if app.aiConfig.AzureOpenAI.APIKey != "" {
+		// Use API key for authentication
+		client := openai.NewClient(
+			azure.WithAPIKey(app.aiConfig.AzureOpenAI.APIKey),
+			azure.WithEndpoint(app.aiConfig.AzureOpenAI.Endpoint, app.aiConfig.AzureOpenAI.APIVersion),
+		)
+		app.openAIClient = client
+		return nil
+	}
```
zhenlan: Nothing wrong with current code but if we can use the following flow it will make the code easier to follow.

```
if ... {
}
else {
}
return nil
```

examples/Go/ChatApp/main.go
```diff
@@ -0,0 +1,232 @@
...
+
+	if len(completion.Choices) == 0 {
+		return "", fmt.Errorf("chat completion returned zero choices")
+	}
+
+	return completion.Choices[0].Message.Content, nil
+}
+
+func (app *ChatApp) runInteractiveChat() {
```
zhenlan: How about this?
```suggestion
func (app *ChatApp) start() {
```

examples/Go/ChatApp/main.go
```diff
@@ -0,0 +1,232 @@
...
+	client := openai.NewClient(
+		azure.WithEndpoint(app.aiConfig.AzureOpenAI.Endpoint, app.aiConfig.AzureOpenAI.APIVersion),
+		azure.WithTokenCredential(tokenCredential),
+	)
+
+	app.openAIClient = client
+	return nil
+}
+
+func (app *ChatApp) callAzureOpenAI(chatConversation []openai.ChatCompletionMessageParamUnion) (string, error) {
```
zhenlan: How about this?
```suggestion
func (app *ChatApp) getAIResponse(chatConversation []openai.ChatCompletionMessageParamUnion) (string, error) {
```

examples/Go/ChatApp/main.go
```diff
@@ -0,0 +1,232 @@
...
+
+	app := &ChatApp{
+		configProvider: configProvider,
+		aiConfig:       aiConfig,
+	}
+
+	// Initialize Azure OpenAI client
+	if err := app.createAzureOpenAIClient(); err != nil {
+		log.Fatalf("Failed to create Azure OpenAI client: %v", err)
+	}
```
zhenlan: We have many nested calls, which makes this sample a little harder to follow. 

It's not clear why we need to create the OpenAI client here. I would move this code into the `app.start()` and use the OpenAI client as a local variable. It doesn't need to be a property of `ChatApp`.

Actually, why do we even need the `ChatApp`? Can we do everything in `app.start()` in `main` directly? I hope we can reduce the amount of code. It will be a lot of code to add in document.

In addition, can we adjust the order of functions? For example, we have the main function first, followed by child functions that are called in main in sequence. It will be easier to follow as most people will read the sample code top-down.

examples/Go/ChatApp/main.go
```diff
@@ -0,0 +1,232 @@
...
+
+		// Get user input
+		fmt.Print("You: ")
+		userInput, err := reader.ReadString('\n')
+		if err != nil {
+			log.Printf("Error reading input: %v", err)
+			continue
+		}
+
+		userInput = strings.TrimSpace(userInput)
```
zhenlan: You have a lot of edge case handling or error handling code throughout this sample. Nothing wrong with them and it's nice to have, but this also makes the code longer. I'd like to minimize the code added to the document. You can either clean them up, or you keep them here but later you only include those lines of code and error handling that are critical when you update the document.

examples/Go/ChatApp/main.go
linglingye001: This refresh is for key values, not key vault. Are you saying we don't need to configure refresh for key values?

examples/Go/ChatApp/main.go
linglingye001: ok, I simplified the code and removed `ChatApp` struct, put most implementation in `main()` with the following order

```go
func main() {
loadAzureAppConfiguration()

initOpenAIClient()

chat conversation loop

...

}

func loadAzureAppConfiguration()

func getAIResponse()

```

examples/Go/ChatApp/main.go
linglingye001: Agree, updated

examples/Go/ChatApp/main.go
zhenlan: Oops, sorry, I mis-read it as Key Vault. Right, don't configure the "interval" and use the default instead.

linglingye001: It's okay :) updated

PR #1095, opened 2025-08-28
Release note for .NET FM 4.3.0

zhiyuanliang-ms:


jimmyca15:
Looks good but given there is no usage instructions in the linked PR, it's a bit hard to discover until we get the feature reference updated. Let's get that out too.

zhiyuanliang-ms:
@jimmyca15
Doc PR:  https://github.com/MicrosoftDocs/azure-docs-pr/pull/304905

PR #1119, opened 2025-12-19
Add request scoping samples for AFD

avanigupta:
Adding some samples to help users set up the right AFD filters according to their application code.

### DIFF
--- a/docs/AzureFrontDoor/readme.md
+++ b/docs/AzureFrontDoor/readme.md
@@ -4,9 +4,9 @@ App Configuration gives developers a single, consistent place to define configur
 
 ## Documentation links
 
-- Conceptual overview of [Hyperscale Configuration](https://learn.microsoft.com/azure/azure-app-configuration/concept-hyperscale-client-config)
+- Conceptual overview of [Hyperscale Configuration](https://learn.microsoft.com/azure/azure-app-configuration/concept-hyperscale-client-configuration)
 - [Connect App Config to Azure Front Door](https://learn.microsoft.com/azure/azure-app-configuration/how-to-connect-azure-front-door)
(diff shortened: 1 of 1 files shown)


docs/AzureFrontDoor/readme.md
```diff
@@ -0,0 +1,265 @@
...
+- [Application uses feature flags only](#application-uses-feature-flags-only)
+- [Application uses key-values and feature flags](#application-uses-key-values-and-feature-flags)
+- [Application uses multiple key-value selectors](#application-uses-multiple-key-value-selectors)
+- [Application uses key-values and snapshot selectors](#application-uses-key-values-and-snapshot-selectors)
+- [Application uses snapshot reference](#application-uses-snapshot-reference)
+- [Application loads key-values with reserved characters](#application-loads-key-values-with-reserved-characters)
+- [Application loads key-values with tag filters](#application-loads-key-values-with-tag-filters)
+
+
+### Application uses key-values only
```
jimmyca15: I think we should start with 

```
builder.Configuration.AddAzureAppConfiguration(options =>
{
    options.ConnectAzureFrontDoor(new Uri("{YOUR-AFD-ENDPOINT}"))
            .ConfigureRefresh(refreshOptions =>
            {
                refreshOptions.RegisterAll()
                    .SetRefreshInterval(TimeSpan.FromMinutes(1));
            });
});
```

Then the next sample can discuss if a `Select` call is used.

jimmyca15: Similarly, with feature flags, I think we should start just with `UseFeatureFlags` being called, then the following sample could discuss what to do if a feature flag selector is used.

avanigupta: Updated, thanks!

jimmyca15 (approved):

PR #1147, opened 2026-04-09
Add .NET sample chat agent app

MaryanneNjeri:
This PR

Adds a .NET sample chat agent application that utilizes the Agent framework


examples/DotNetCore/ChatAgent/Program.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        Uri endpoint = new(Environment.GetEnvironmentVariable("AZURE_APPCONFIGURATION_ENDPOINT") ??
+            throw new InvalidOperationException("The environment variable 'AZURE_APPCONFIGURATION_ENDPOINT' is not set or is empty"));
+        options.Connect(endpoint, credential)
+            .Select("ChatAgent:*")
+            .ConfigureRefresh(refreshOptions =>
+            {
+                refreshOptions.RegisterAll();
+            });
+        refresher = options.GetRefresher();
```
jimmyca15: refresher is unused

jimmyca15: Did you intend to use it? It will require recreating the agent which we may want to leave out. I see that in Python we don't have refresh, so perhaps it should be omitted here.

MaryanneNjeri: Didn't intend to use refresher. Updated to remove refresher

jimmyca15 (approved):

